PR #1, opened 2016-05-09
add constraints for CIDR variables, and selectable instances types with low/medium/high/10gbit networking for NAT instances

cnelson:
There are now constraints for CIDR variables, and selectable instances types with low/medium/high/10gbit networking for NAT instances

### DIFF
--- a/update.sh
+++ b/update.sh
@@ -1,0 +1,123 @@
+#!/bin/bash
+
+if [ $# -eq 0 ] || \
+   [ "$AWS_ACCESS_KEY_ID" == "" ] || \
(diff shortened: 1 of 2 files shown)


update.sh
```diff
@@ -0,0 +1,123 @@
...
+##### SETTINGS
+
+# the name of our stack
+STACK_NAME="$1"
+
+# s3 urls
+S3_BASE="https://s3-${AWS_DEFAULT_REGION}.amazonaws.com"
+
+# the bucket
+BUCKET="cf-templates-1v2czdpr8i6tw-us-gov-west-1"
```
LinuxBozo: bucket probably should be definable as well instead of hardcoded

PR #7, opened 2016-05-25
Add the bootstrap stack

cnelson:
Add a bootstrap stack which provisions a single ec2 instance in the default VPC running concourse and starts the the pipeline you see in this PR.

That pipeline runs terraform apply using instance profiles for AWS access and deploys the `test` stack successfully (The test stack just creates an empty security group to prove terraform is working).

@LinuxBozo if this looks good to you, I should be able to have it running real stacks tomorrow.  When I start on that, is `cloud-gov` the stack I should be applying? is that the top level?

### DIFF
--- a/terraform/modules/bootstrap_concourse/sg_bootstrap.tf
+++ b/terraform/modules/bootstrap_concourse/sg_bootstrap.tf
@@ -1,14 +1,7 @@
 resource "aws_security_group" "bootstrap" {
   ingress {
-    from_port = 22
-    to_port = 22
-    protocol = "tcp"
-    cidr_blocks = ["0.0.0.0/0"]
-  }
-
-  ingress {
-    from_port = 80
-    to_port = 80
+    from_port = 443
+    to_port = 443
     protocol = "tcp"
     cidr_blocks = ["0.0.0.0/0"]
   }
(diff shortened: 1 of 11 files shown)


terraform/stacks/bootstrap/variables.tf
```diff
@@ -0,0 +1,19 @@
+variable "concourse_password" {}
+
+variable "default_region" {}
```
LinuxBozo: Any reason to not just default this to `us-gov-west-1`? Especially since you provide a default in the `bootstrap_concourse` module itself

terraform/modules/bootstrap_concourse/sg_bootstrap.tf
```diff
@@ -0,0 +1,22 @@
+resource "aws_security_group" "bootstrap" {
+  ingress {
+    from_port = 22
+    to_port = 22
```
LinuxBozo: Do we really need to be able to ssh in to the concourse box?

cnelson: I added this as a debugging option since this thing creates the world, if something goes wrong you might want to be able to find out why, especially if something goes wrong before concourse runs your job and you have nothing to hijack.

However I haven't actually needed this yet in my testing, so I can remove it if you feel strongly about it.

LinuxBozo: I think this is definitely a YAGNI. The only reason you'd need it is if Concourse itself doesn't come up, but then that points to a problem with the AMI, which means digging into the packer setup.

PR #14, opened 2016-06-07
[WIP] Terraform to create ELBs for Cloudfoundry.  

cnelson:
@dlapiduz / @LinuxBozo is this what you were expecting? 

At minimum this probably needs security groups added to it before merge, but I wasn't sure what those should be.  The ELB configuration was copied from https://github.com/18F/cg-manifests/blob/0a597f2617d67096b66ded9e6ee007a65f2fb6d3/provisioning/elb.json but I have no idea how far out of sync that is with what you are actually running in us-east.

Feedback appreciated.

This is for 18F/cg-atlas#64

### DIFF
--- a/terraform/stacks/production/stack.tf
+++ b/terraform/stacks/production/stack.tf
@@ -23,6 +23,6 @@ module "cf" {
     stack_description = "${var.stack_description}"
     elb_main_cert_name = "${var.main_cert_name}"
     elb_apps_cert_name = "${var.apps_cert_name}"
-    elb_subnets = "${module.stack.public_subnet_az1},${module.stack.public_subnet_az1}"
+    elb_subnets = "${module.stack.public_subnet_az1},${module.stack.public_subnet_az2}"
     elb_security_groups = "${module.stack.web_traffic_security_group}"
 }
--- a/terraform/stacks/production/stack.tf
+++ b/terraform/stacks/production/stack.tf
@@ -15,3 +15,14 @@ module "stack" {
     target_stack_name = "${var.target_stack_name}"
 
 }
+
+module "cf" {
+    source = "../../modules/cloudfoundry"
+
+    account_id = "${var.account_id}"
+    stack_description = "${var.stack_description}"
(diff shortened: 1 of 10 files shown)


dlapiduz:
I havent been doing terraform in a while so I will leave it to @LinuxBozo

LinuxBozo:
@cnelson The security group list will only have web traffic one right now, so:
```
elb_security_groups = "${module.stack.web_traffic_security_group}"
```
Also, missing a listener for port 4443 -> 80  TCP-SSL

terraform/stacks/production/stack.tf
```diff
@@ -14,4 +14,15 @@ module "stack" {
...
+}
+
+module "cf" {
+    source = "../../modules/cloudfoundry"
+
+    account_id = "${var.account_id}"
+    stack_description = "${var.stack_description}"
+    elb_main_cert_name = "${var.main_cert_name}"
+    elb_apps_cert_name = "${var.apps_cert_name}"
+    elb_subnets = "${module.stack.public_subnet_az1},${module.stack.public_subnet_az1}"
```
LinuxBozo: These are both az1, second should be az2

terraform/stacks/staging/stack.tf
```diff
@@ -14,4 +14,15 @@ module "stack" {
...
+}
+
+module "cf" {
+    source = "../../modules/cloudfoundry"
+
+    account_id = "${var.account_id}"
+    stack_description = "${var.stack_description}"
+    elb_main_cert_name = "${var.main_cert_name}"
+    elb_apps_cert_name = "${var.apps_cert_name}"
+    elb_subnets = "${module.stack.public_subnet_az1},${module.stack.public_subnet_az1}"
```
LinuxBozo: Same issue as above, az2 missing

cnelson:
I created a story 18F/cg-atlas#66 to add the custom SSL negotiation policy as soon as terraform supports doing so.

cnelson:
Update the stage and prod stacks to use the *.fr certs for the app ELB as well for now per @dlapiduz

PR #24, opened 2016-06-10
Aws broker user

jmcarp:
The aws broker pipeline needs to provision a handful of rds instances to store internal state and support shared services. Once deployed, the broker app needs to create, describe, and delete rds instances. Does this look right @LinuxBozo?


LinuxBozo:
@jmcarp on the broker access to the terraform state bucket, if it's using its own stack to manage state, and read state from parent stacks (for things like networking settings), I would split the S3 access a bit.. I would give read access (GetObject, ListBucket) to the overall state bucket, and limit PutObject to the broker stack subdir in the bucket so it can't write it's parent stack state.

PR #25, opened 2016-06-10
Drop account ids from s3 resources.

jmcarp:
One day I'm going to get this right on the first try.


LinuxBozo:
@jmcarp you and me both.

PR #35, opened 2016-06-16
Add VPC flow logs to bosh_vpc

dlapiduz:


terraform/modules/bosh_vpc/vpc.tf
```diff
@@ -13,4 +13,59 @@ resource "aws_vpc" "main_vpc" {
...
+    name = "${var.stack_description}-flow-log-role"
+    assume_role_policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Sid": "",
+      "Effect": "Allow",
+      "Principal": {
+        "Service": "vpc-flow-logs.amazonaws.com"
```
LinuxBozo: @dlapiduz this isn't some odd thing that may be different for govcloud is it? Also, if it is, when you change it, make sure you leave a TODO in a comment to come back and make this configurable.

dlapiduz: I cant find it where its different but I see why you ask... I have no idea

LinuxBozo: Yeah, can't find anything about it either. So :shipit:

PR #37, opened 2016-06-29
Provision s3 bucket and iam user for cdn-broker.

jmcarp:
Policies from https://github.com/18F/cf-cdn-service-broker/issues/19.

* Create publicly readable bucket for LE verification tokens
* Create IAM user that can read/write bucket, IAM certs, and cloudfront distributions

Note--I didn't run this locally, since I'm not sure the best way to do it, short of grabbing all the envvars from concourse and setting them locally. Could use a review from @LinuxBozo at some point. Not urgent--I really just wanted to knock this out before I forgot which policies the cdn broker needed.


LinuxBozo:
Won't this broker be run possibly in a CF related stack, which means, shouldn't be attached to tooling stack, but the staging/prod stacks?

jmcarp:
Yeah, that seems right--I just threw this stuff in tooling because we already had users for other service brokers there already--which I might've gotten wrong when I added them earlier. Do you think we should have separate IAM users and S3 buckets for staging and prod? I can move all the broker users/buckets over if so.

dlapiduz:
One thing about this... we shouldn't run it in govcloud since there is no CloudFront there... we should create a separate Terraform script/stack and run it in an e/w environment

jmcarp:
Derp, that's right. So do we want a separate stack that runs in e/w? I can move this stuff into terraform/stacks/ew (or whatever we should call it) and add a separate ci job for it.

jmcarp:
@LinuxBozo: what do you think about moving the bucket and iam user to a separate stack that runs in east/west, like `stacks/tooling-external.tf`? Like you were saying earlier, these resources could also go into staging and production cf stacks (maybe `stacks/staging-external` and `stacks/production-external`), but so far we haven't been creating separate broker users per environment. We can do that, either here or in another task.

LinuxBozo:
@jmcarp I would say that since we're storing LE certs here, we should limit exposure and do per environment users. You could create this as a module then, and just include in any top level stack, like staging or production, or use the module in it's own bit of terraform to backfill something like E/W.

jmcarp:
@LinuxBozo: updated to create an "external" stack that's run for both staging and production environments. I put this in its own stack because it needs to be run with its own aws creds, region, and partition. Also, this now includes #46.

PR #45, opened 2016-07-11
Parameterize AWS partition.

jmcarp:
Knock out TODO items related to AWS partition/path.

@LinuxBozo


terraform/modules/iam_user/cf_user/user.tf
```diff
@@ -4,7 +4,6 @@ module "cf_user" {
 
     username = "${var.username}"
 
-    /* TODO: Make the bucket names configurable */
     /* TODO: Make sure the bucket arn:path is configurable */
```
LinuxBozo: Removed wrong TODO?

jmcarp: Ack, fixed.

PR #46, opened 2016-07-14
Reject PUT without encryption.

jmcarp:
Part of https://github.com/18F/cg-atlas/issues/123.

@LinuxBozo: I went with a custom module here to avoid copying and pasting bucket policies, but it occurred to me that this won't be composable with other bucket policies--maybe I should embrace the verbosity of terraform and revert the module. What's your preference?

Note: this should be safe to merge after https://ci.fr.cloud.gov/pipelines/deploy-cf/jobs/deploy-cf-staging/builds/109 and the corresponding production deploy pass, now that they're applying https://github.com/18F/cg-deploy-cf/commit/a7405f19ab2b680f11691e88cd8dceec09cd2a85.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,6 @@ Manual steps to create the world:
 1. Make sure you have [terraform](https://www.terraform.io/), awscli, and jq installed
  1. Mac OSX users can install all of these with homebrew
 1. Create S3 bucket with versioning enabled to store terraform state: `terraform-state`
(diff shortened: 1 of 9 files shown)


jmcarp:
@LinuxBozo: updated to provision encrypted and/or public buckets on bootstrap, although some of the buckets are commented out until upstream accepts / releases our changes.

PR #48, opened 2016-07-18
Add prevent destroy lifecycle flag.

jmcarp:
To be used for databases that shouldn't be automatically deleted on
configuration changes.

Specifically, I noticed that the rds instances in the aws broker would have been silently destroyed if we had re-run the concourse pipeline after switching the default encryption from false to true.

@LinuxBozo

### DIFF
--- a/terraform/modules/rds/database.tf
+++ b/terraform/modules/rds/database.tf
@@ -21,6 +21,7 @@ resource "aws_db_instance" "rds_database" {
 
   lifecycle {
     ignore_changes = ["identifier"]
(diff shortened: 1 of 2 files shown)


LinuxBozo:
Is this necessary? Couldn't you have updated your terraform for the broker to just set encryption to false instead so it matched?

jmcarp:
If a build of aws-broker had been triggered (for example, if somebody had updated the code), then the pipeline would've triggered the terraform apply script and silently wiped out the data. We could have made a note to set the encryption flag on aws-broker after changing the default here, but that seems like an easy thing to forget.

PR #50, opened 2016-07-19
Update cloudwatch resource.

jmcarp:
Limiting the `cloudwatch*` actions to the `logs` resource means we can't access metrics, so the policy needs to be updated. As far as I can tell, the current resource of `arn:${var.aws_partition}:logs:*:*:*` effectively grants access to everything logs-related, so I think that `*` is a reasonable way to allow cloudwatch metrics access without changing behavior. But if not, I'll separate the policy into two statements, one for logs and another for metrics.


LinuxBozo:
@jmcarp given use of specific actions, I think thats fine.

PR #52, opened 2016-07-20
Added read-only-admin account. Pl see https://github.com/18F/cg-atlas…

madhujoshi:
…/issues/70


terraform/modules/iam_user/read_only_admin/user.tf
```diff
@@ -0,0 +1,169 @@
+
+module "read_only_admin" {
+    source = ".."
+
+    username = "${var.username}"
+
+    /* TODO: Make the bucket names configurable */
```
jmcarp: Looks like we don't refer to bucket names or `subgrp` here. Are these TODOs relevant?

terraform/modules/iam_user/read_only_admin/variables.tf
```diff
@@ -0,0 +1,5 @@
+variable "username" {}
+variable "account_id" {}
```
jmcarp: Looks like some of these variables aren't used.

terraform/modules/iam_user/read_only_admin/user.tf
```diff
@@ -0,0 +1,169 @@
...
+        {
+            "Sid": "CloudWatchLogsSpecific",
+            "Effect": "Allow",
+            "Action": [
+                "logs:GetLogEvents",
+                "logs:DescribeLogGroups",
+                "logs:DescribeLogStreams"
+            ],
+            "Resource": [
+                "arn:aws:logs:*:*:*"
```
jmcarp: We shouldn't hard-code partitions like `aws`, since this would be `aws-us-gov` in govcloud. We've been using the `aws_partition` variable to handle this elsewhere. Also, just curious--why does cloudwatch go in its own statement?

madhujoshi:
Thanks Josh.  I've removed the extra variables and cleaned up the comment section.

PR #53, opened 2016-07-20
Added TF module for Diego

frsfx:


terraform/modules/diego/sg_mysql.tf
```diff
@@ -0,0 +1,30 @@
...
+
+resource "aws_security_group" "diego_rds_mysql" {
+  description = "Allow access to incoming mysql traffic"
+  vpc_id = "${var.vpc_id}"
+
+  ingress {
+    from_port = 3306
+    to_port = 3306
+    protocol = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
```
dlapiduz: @frsfx can we restrict the incoming traffic to something?

frsfx:
@LinuxBozo I updated files yesterday based on the changes we discussed over slack

terraform/modules/diego/diego_elb_main.tf
```diff
@@ -0,0 +1,56 @@
+/*
+ * Variables required:
+ *   stack_description
+ *   elb_subnets
+ *   
+ */
+
+
+resource "aws_security_group" "diego-elb-sg" {
+  name = "diego-elb-sg"
```
LinuxBozo: Hard coding this name will cause an issue once you run from production stack. May want to use `stack_description` in here to make it unique.

terraform/modules/diego/outputs.tf
```diff
@@ -0,0 +1,19 @@
...
+output "elb_diego_main_name" {
+  value = "${aws_elb.diego_elb_main.name}"
+}
+
+output "elb_diego_main_dns_name" {
+  value = "${aws_elb.diego_elb_main.dns_name}"
+}
+
+/* Diego subnets */
+output "services_subnet_az1" {
```
LinuxBozo: May want to change these output names to make it clear it's for diego

LinuxBozo:
@frsfx You'll want to update `variables.tf` in the staging stack for the diego cidr variables and pass along the outputs from your diego module in `outputs.tf`. Also, you'll want to update the `ci/pipeline.yml` staging job to pass along your diego cidrs using the appropriate TF_VARs.

terraform/stacks/staging/outputs.tf
```diff
@@ -125,3 +125,22 @@ output "services_subnet_az1" {
...
+output "diego_elb_dns_name" {
+  value = "${module.diego.diego_elb_main.dns_name}"
+}
+
+/* Diego subnets */
+output "diego_services_subnet_az1" {
+  value = "${module.diego.diego_az1_services.id}"
+}
+output "diego_services_subnet_az2" {
+  value = "${module.diego.diego_az2_services.id}"
```
LinuxBozo: When passing on outputs from a module, you use the module output reference, so simply `module.diego.diego_services_subnet_az2`  Same goes for the rest of these added outputs.

PR #55, opened 2016-07-20
Temporarily revert prevent_destroy.

jmcarp:
Combining `prevent_destroy` with `ignore_changes` on a dynamic variable
effectively makes it impossible to update the relevant resource, as
described in https://github.com/hashicorp/terraform/issues/7728. This
patch reverts the `prevent_destroy` flag on the rds resource. If the
underlying issue is fixed, we can restore this flag. Otherwise, we'll
have to either specify all database identifiers or allow them to be
randomly generated.

@LinuxBozo

### DIFF
--- a/terraform/modules/rds/database.tf
+++ b/terraform/modules/rds/database.tf
@@ -21,7 +21,8 @@ resource "aws_db_instance" "rds_database" {
 
   lifecycle {
     ignore_changes = ["identifier"]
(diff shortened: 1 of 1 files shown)


LinuxBozo:
@jmcarp Time to revisit.. Looks like they fixed it according to Mitchell https://github.com/hashicorp/terraform/issues/7728#issuecomment-258066903

PR #61, opened 2016-08-02
Add private S3 endpoint to VPCs

LinuxBozo:
@jmcarp


terraform/modules/bosh_vpc/endpoint_s3.tf
```diff
@@ -0,0 +1,17 @@
+resource "aws_vpc_endpoint" "private-s3" {
+    vpc_id = "${aws_vpc.main_vpc.id}"
+    service_name = "com.amazonaws.${var.aws_default_region}.s3"
+    route_table_ids = ["${aws_route_table.az1_private_route_table.id}", "${aws_route_table.az1_private_route_table.id}"]
```
jmcarp: Should the second value be `"${aws_route_table.az2_private_route_table.id}"`?

LinuxBozo: Good catch. Fixed.

PR #112, opened 2016-10-21
Update health_check to be shorter than 5 minutes

rogeruiz:


rogeruiz:
Thanks @datn.

These changes are to shorten the `health_check` in terms of `health_threshold` and `interval`, but not `timeout` that one got doubled. 👍

PR #118, opened 2016-10-27
Create totp seed table in uaadb for MFA

LinuxBozo:


ci/scripts/create-and-update-db.sh
```diff
@@ -21,4 +21,9 @@ for db in ${DATABASES}; do
     psql_adm -d "${db}" -c "CREATE EXTENSION IF NOT EXISTS \"${ext}\""
   done
 
+  # Special case for uaadb, create totp seed table for use with MFA
+  if [ "${db}" = "uaadb" ]; then
+    psql_adm -d "${db}" -c "CREATE TABLE IF NOT EXISTS totp_seed ( username varchar(255) PRIMARY KEY, seed varchar(36), backup_code varchar(36) )"
```
rogeruiz: Curiosity. Are we only going only going to be storing one backup code? My familiarity with back-up codes is that they'd be a list. Similar to Github's and Google's backup code. Either way though, this is good for now. 👍

LinuxBozo: There really is no reason to have more than one. Multiple backup codes has been discovered to be completely confusing. If one works, why do I need multiple? In case I forgot that one? But if I forget that one, because I didn't save it, what makes you think I saved the other 9?

rogeruiz: 😂 👏

PR #149, opened 2016-12-06
Enable encryption on blobstore buckets.

jmcarp:
Resolves https://github.com/18F/cg-atlas/issues/128.


jmcarp:
WDYT @LinuxBozo? Thoughts on how to verify this?

jmcarp:
Yep--this might be a fun opportunity to try out `terraform import`. The only reason I haven't done this already is because the buckets get created during bootstrap, which afaict isn't going to run until the next time we set up a new environment. Which is why I'm not sure how to verify manually 😖

PR #151, opened 2016-12-17
Add golang s3 broker user.

jmcarp:
Based on https://github.com/cloudfoundry-community/s3-broker/blob/master/iam_policy.json. Also, pretty neat to get json linting with the `template_file` resource.

cc @cnelson


cnelson:
Why this versus the existing s3 broker user?  If the concern is sharing creds between brokers, I'd hope that would be temporary as we'd kill the java broker as soon as possible after the golang one is live.

jmcarp:
The two brokers require slightly different permissions, so I was thinking it would be simpler to keep the iam users separate instead of adding the new permissions to the existing user, then dropping the unused permissions later. Either way is fine with me, wdyt?

PR #153, opened 2016-12-23
Create trigger for verifying cloud.gov origin

rogeruiz:
Thanks to @cnelson's advice on using Triggers, we can now verify that whenever we update a user in the DB we'll enforce that cloud.gov users are properly set up.

I verified that the `verified` is set to false and that `created` and `passwd_lastmodified` match by inviting a test user. I think this might be all we need. But I'd appreciate extra eyes on this.

cc: @linuxbozo

### DIFF
--- a/ci/scripts/create-and-update-db.sh
+++ b/ci/scripts/create-and-update-db.sh
@@ -26,7 +26,7 @@ for db in ${DATABASES}; do
   # Special case for Shibboleth, create function and trigger that verifies origin uaa is set to cloud.gov IdP
   # Special case for Shibboelth, create FK between totp_seed and users and CASCADE on delete
   if [ "${db}" = "uaadb" ]; then
-    psql_adm -d "${db}" -c << EOT
+    psql_adm -d "${db}" <<-EOT
     CREATE TABLE IF NOT EXISTS totp_seed
       (
         username varchar(255) PRIMARY KEY,
@@ -34,7 +34,7 @@ for db in ${DATABASES}; do
         backup_code varchar(36)
       )
 EOT
-    psql_adm -d "${db}" -c << EOT
+    psql_adm -d "${db}" <<-EOT
       ALTER TABLE IF EXISTS totp_seed
         ADD CONSTRAINT username_record_keeper
           FOREIGN KEY (username)
@@ -42,7 +42,7 @@ EOT
           ON DELETE CASCADE
 EOT
 
-    psql_adm -d "${db}" -c << EOT
+    psql_adm -d "${db}" <<-EOT
       CREATE TABLE IF NOT EXISTS storagerecords
         (
           context varchar(255) NOT NULL,
@@ -54,12 +54,12 @@ EOT
         )
 EOT
 
-    psql_adm -d "${db}" -c << EOT
+    psql_adm -d "${db}" <<-EOT
       CREATE OR REPLACE FUNCTION "f_isValidEmail"( text ) RETURNS BOOLEAN AS '
       SELECT $1 ~ ''^[^@\s]+@[^@\s]+(\.[^@\s]+)+$'' AS RESULT
       ' LANGUAGE sql
 EOT
-    psql_adm -d "${db}" -c << EOT
+    psql_adm -d "${db}" <<-EOT
       CREATE OR REPLACE FUNCTION "f_enforceCloudGovOrigin"( text ) RETURNS TRIGGER AS $$
       BEGIN
         UPDATE users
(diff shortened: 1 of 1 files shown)


ci/scripts/create-and-update-db.sh
```diff
@@ -23,9 +23,33 @@ for db in ${DATABASES}; do
...
+      ' LANGUAGE sql
+EOT
+    psql_adm -d "${db}" -c << EOT
+      CREATE OR REPLACE FUNCTION "f_enforceCloudGovOrigin"( text ) RETURNS TRIGGER AS $$
+      BEGIN
+        UPDATE users
+          SET ( origin, externalId ) = ( 'cloud.gov', username )
+          WHERE "f_isValidEmail"( username ) AND
+            origin = 'uaa' AND
+            verified = false AND
```
LinuxBozo: Shouldn't this be:
```
verified = true AND
created::date != passwd_lastmodified::date
```
That would indicate that the user has accepted the invite after those two conditions, yeah?

rogeruiz: 🤦‍♂️ YES

rogeruiz:
This should be good for now, but I have some local changes I'd like to integrate for the `FK` and `CASCADE` stuff we discussed.

ci/scripts/create-and-update-db.sh
```diff
@@ -23,9 +23,33 @@ for db in ${DATABASES}; do
...
+          WHERE "f_isValidEmail"( username ) AND
+            origin = 'uaa' AND
+            verified = true AND
+            created::date != passwd_lastmodified::date;
+      END;
+      $$ LANGUAGE plpgsql
+EOT
+    psql_adm -d "${db}" -c << EOT
+      CREATE TRIGGER enforce_cloud_gov_idp_origin_trigger
+        AFTER INSERT OR UPDATE ON users
```
cnelson: Can this be done just on UPDATE?  since UAA inserts the user when it invites (and we never want to do anything then), and then presumably does an UPDATE to set their password / verified flag which is when we want to catch this?

rogeruiz:
All the Heredocs here are wrong. Or at least they don't work locally. I had been testing this within the DB client. :frown:

rogeruiz:
Chips away at 18F/cg-product#255

ci/scripts/create-and-update-db.sh
```diff
@@ -23,9 +23,63 @@ for db in ${DATABASES}; do
...
+    psql_adm -d "${db}" <<-EOT
+    CREATE TABLE IF NOT EXISTS totp_seed
+      (
+        username varchar(255) PRIMARY KEY,
+        seed varchar(36),
+        backup_code varchar(36)
+      )
+EOT
+    psql_adm -d "${db}" <<-EOT
+      ALTER TABLE IF EXISTS totp_seed
```
jmcarp: Is it possible for this table not to exist at this point?

ci/scripts/create-and-update-db.sh
```diff
@@ -23,9 +23,63 @@ for db in ${DATABASES}; do
...
+        )
+EOT
+
+    psql_adm -d "${db}" <<-EOT
+      CREATE OR REPLACE FUNCTION "f_isValidEmail"( text ) RETURNS BOOLEAN AS '
+      SELECT $1 ~ ''^[^@\s]+@[^@\s]+(\.[^@\s]+)+$'' AS RESULT
+      ' LANGUAGE sql
+EOT
+    psql_adm -d "${db}" <<-EOT
+      CREATE OR REPLACE FUNCTION "f_enforceCloudGovOrigin"( text ) RETURNS TRIGGER AS $$
```
jmcarp: Your call on this, since I don't know the UAA internals, but some inline docs might be useful here. For example, it's not obvious to me why we need the email check, or what exactly the date comparison does.

ci/scripts/create-and-update-db.sh
jmcarp: Another idempotency question--do we need to drop the constraint if it exists and re-add? I *think* we do, but haven't verified.

ci/scripts/create-and-update-db.sh
rogeruiz: thanks, added a comment above the `psql_adm` command.

ci/scripts/create-and-update-db.sh
rogeruiz: Yeah good call.

rogeruiz: Not any more but sure. Also, it's certainly safer to make it all happen within a single transaction.

PR #154, opened 2016-12-26
add a user that has read-only access to cloud-gov-bosh-releases 

cnelson:
so the east deployment can consume them


jmcarp:
LGTM, should we also add outputs to tooling/outputs.tf for the new user?

cnelson:
Good call, I always forget to do that.  updated!

PR #159, opened 2017-01-06
Add permissions for rds broker acceptance tests.

jmcarp:
Alternatively, since these permissions are only needed for running the acceptance tests, we could make a separate user for tests and leave this permissions out for the broker user. WDYT?

### DIFF
--- a/terraform/modules/iam_user/aws_broker_user/user.tf
+++ b/terraform/modules/iam_user/aws_broker_user/user.tf
@@ -19,10 +19,13 @@ module "aws_broker_user" {
                 "rds:ModifyDBInstance",
                 "rds:AddTagsToResource",
                 "rds:ListTagsForResource",
-                "rds:RemoveTagsFromResource"
+                "rds:RemoveTagsFromResource",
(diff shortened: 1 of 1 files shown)


jmcarp:
The gds broker saves a snapshot on delete by default, but can be configured not to. The relevant integration tests verify that the snapshot exists in the first case but not in the second: https://github.com/alphagov/paas-cf/blob/master/platform-tests/src/acceptance/rds_broker_test.go#L147-L187. The broker doesn't need the permissions because it's allowed to pass the snapshot-on-delete flag to the delete api call (apparently).

LinuxBozo:
Is there some other downside of giving the broker user this perm that I'm not seeing? Is there any plan in the future to give the broker a proper dashboard where snapshots could be managed by the user? (just spitballing here)

jmcarp:
No, I think this is fine--the only possible problem I can think of is that we mess up the broker code and it deletes snapshots by mistake, but there's no code in the broker that touches snapshots at all, so I'm not too concerned.

PR #160, opened 2017-01-09
Little patches to fix the uaadb

rogeruiz:
cc: @linuxbozo

### DIFF
--- a/ci/scripts/create-and-update-db.sh
+++ b/ci/scripts/create-and-update-db.sh
@@ -34,8 +34,9 @@ for db in ${DATABASES}; do
           seed varchar(36),
           backup_code varchar(36)
         );
-      ALTER TABLE IF EXISTS totp_seed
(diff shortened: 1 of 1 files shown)


ci/scripts/create-and-update-db.sh
```diff
@@ -34,8 +34,9 @@ for db in ${DATABASES}; do
           seed varchar(36),
           backup_code varchar(36)
         );
-      ALTER TABLE IF EXISTS totp_seed
-        DROP CONSTRAINT username_record_keeper;
+      IF EXISTS (SELECT 1 from sys.objects where name = 'username_record_keeper')
+        ALTER TABLE IF EXISTS totp_seed
+          DROP CONSTRAINT username_record_keeper;
```
rogeruiz: I noticed that this was failing before because it couldn't drop a constraint that didn't exist.

PR #162, opened 2017-01-12
Grant riemann the ability to snapshot and stop VMs

cnelson:
This is for https://github.com/18F/cg-product/issues/476


terraform/modules/iam_role/riemann_monitoring/role.tf
```diff
@@ -0,0 +1,44 @@
+module "riemann_monitoring" {
+  source = ".."
+
+  role_name = "${var.role_name}"
+  iam_policy = <<EOF
+{
```
jmcarp: WDYT about moving this into a separate template file while we're at it? I'm tired of committing policies with invalid JSON 😄

cnelson:
Moved the policy to a separate file

terraform/modules/iam_role/riemann_monitoring/policy.json
```diff
@@ -0,0 +1,35 @@
...
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "logs:DescribeLogGroups",
+        "logs:DescribeLogStreams"
+      ],
+      "Resource": [
+        "arn:${var.aws_partition}:logs:${var.aws_default_region}:${var.account_id}:*"
```
jmcarp: For some reason, we don't need the `var.` prefix in templates: https://www.terraform.io/docs/providers/template/d/file.html

cnelson: prefixes removed

PR #164, opened 2017-01-23
Add environment blobstores

rogeruiz:
Tied to the AC from 18F/cg-atlas#128

I verified that the v260.4 boshrelease has been deployed and that the s3-resource changes had been merged.

cc: @linuxbozo, @jmcarp


jmcarp:
Is this different from #149?

rogeruiz:
@jmcarp it's worse. 🤦‍♂️ thanks!

PR #168, opened 2017-02-22
Save terraform yaml to s3.

jmcarp:
h/t @cnelson

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -68,6 +68,13 @@ jobs:
     params:
       <<: *external-staging-params
       TERRAFORM_ACTION: apply
+  - task: terraform-state-to-yaml
+    file: pipeline-tasks/terraform-state-to-yaml.yml
+    params:
+      STATE_FILE: terraform.tfstate
+  - put: terraform-yaml-external-staging
+    params:
+      file: terraform-yaml/state.yml
 
 - name: plan-bootstrap-external-production
   plan:
@@ -107,6 +114,13 @@ jobs:
     params:
       <<: *external-production-params
       TERRAFORM_ACTION: apply
+  - task: terraform-state-to-yaml
+    file: pipeline-tasks/terraform-state-to-yaml.yml
+    params:
+      STATE_FILE: terraform.tfstate
+  - put: terraform-yaml-external-production
+    params:
+      file: terraform-yaml/state.yml
 
 - name: plan-bootstrap-tooling
   plan:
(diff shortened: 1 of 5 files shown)


ci/pipeline.yml
```diff
@@ -477,6 +512,51 @@ resources:
...
+  type: s3
+  source:
+    bucket: {{aws_external_s3_tfstate_bucket}}
+    versioned_file: {{tf_state_file_external_staging}}
+    access_key_id: {{aws_external_access_key_id}}
+    secret_access_key: {{aws_external_secret_access_key}}
+    region_name: {{aws_external_region}}
+
+- name: terraform-yaml-external-production
+  type: s3
```
cnelson: Can we use s3-iam for all of these?

jmcarp: For some reason, I only let the concourse role write to one file: https://github.com/18F/cg-provision/blob/master/terraform/modules/iam_role/concourse_worker/role.tf#L40-L48

cnelson: WDYT about expanding that to the whole tfstate bucket?

jmcarp: SGTM, updated.

terraform/modules/iam_role/concourse_worker/role.tf
```diff
@@ -43,7 +43,8 @@ module "concourse" {
         "s3:PutObject"
```
cnelson: This is a versioned bucket, I think we need to allow:
s3:PutObject
s3:PutObjectAcl
s3:GetObject
s3:ListBucket
s3:GetBucketVersioning
s3:GetObjectVersion
s3:ListBucketVersions
s3:PutObjectVersionAcl

terraform/modules/iam_role/concourse_worker/role.tf
```diff
@@ -43,7 +43,8 @@ module "concourse" {
...
       ],
       "Resource": [
-        "arn:${var.aws_partition}:s3:::cloud-gov-varz/master-bosh-state.json"
+        "arn:${var.aws_partition}:s3:::cloud-gov-varz/master-bosh-state.json",
+        "arn:${var.aws_partition}:s3:::terraform-state/*"
```
cnelson: and probably on the bucket itself as well `arn:${var.aws_partition}:s3:::terraform-state`.

Should we be passing these bucket names in as variables rather than hardcoding them?

jmcarp:
Updated. The role should already get the get/list actions on the bucket and contents from the previous statement.

PR #180, opened 2017-02-28
Add separate logging load balancer.

jmcarp:
Part of moving doppler / loggregator to 443.


jmcarp:
Note: adding listeners on both 443 and 4443 so that we can point dns to this elb without breaking logs, then switch over logging endpoints in cf.

PR #181, opened 2017-03-01
[ work in progress ] Move the cdn_broker user where it should live

rogeruiz:
This branch moves the CDN Broker user from the main `modules/*` path to `modules/iam_user/*` path. There are going to be some rotations for the `iam_user` modules that has this commit as a parent, since they will be updating the `cdn_broker` user as well. The rotation documentation is being worked on here 18F/cg-site#802 and will reference modifying the `iam_user/*` files to rotate secrets.


jmcarp:
Does it make sense to create an s3 bucket in a module inside `iam_users`?

rogeruiz:
@jmcarp I _think_ so, since this particular user depends on the bucket and it's a referenced in the `cdn_broker` module for `stacks/external/stack.tf`. Having all the users in one place also makes it a bit easier to write the documentation for rotating secrets I'm working on here => 18F/cg-site#802.

jmcarp:
Does the bucket depend on the user? It looks like the bucket name is passed to both.

terraform/modules/iam_user/cdn_broker/resources.tf
```diff
@@ -1,5 +1,5 @@
 module "cdn_broker_bucket" {
```
rogeruiz: So should this be moved this over to `modules/s3_bucket/cdn_broker_bucket`?

terraform/stacks/external/stack.tf
```diff
@@ -1,5 +1,5 @@
 module "cdn_broker" {
```
rogeruiz: And if we move the bucket creation to `modules/s3_bucket/cdn_broker_bucket` how does this module definition change? can we have multiple source? Would there be a `cdn_broker_bucket` module in this file now instead of just a single `module "cdn_broker"`?

LinuxBozo: Yes, there would be two modules included in this stack. One for the user, one for the bucket.

rogeruiz:
This is old and addressed in other PRs.

PR #182, opened 2017-03-02
Simplify iam user key rotation.

jmcarp:
* Add previous and current keys to all iam users
* Move iam policies to template files
* Deprecate iam user module

We'll need to run a bunch of `terraform state mv` commands to apply these changes without destroying and recreating resources.


rogeruiz:
@jmcarp is going to update the TF state and move the creds to the new naming scheme with the pipeline paused. Then we'll be able to start rotating everything in pieces. Great work @jmcarp 👏 Now I just need to update the documentation to be less complicated in the WIP docs.  18F/cg-site#802

PR #183, opened 2017-03-02
Begin rotation of iam credentials for external stack.

jmcarp:
@rogeruiz: we can probably just grab half the diff here for the docs.


rogeruiz:
Great! I grabbed the patch from here. thanks! https://patch-diff.githubusercontent.com/raw/18F/cg-provision/pull/183.patch

PR #186, opened 2017-03-03
Clean up listeners on 4443.

jmcarp:
Don't merge until verifying that *all* firehose consumers are using the correct port!


cnelson:
WDYT about removing port 4443 from the security groups that reference it as well?

jmcarp:
Thanks for catching that. Since we now have no load balancers listening on 4443, I dropped all security group ingress rules for 4443.

PR #188, opened 2017-03-06
Drop postgres 9.4 databases and alias outputs.

jmcarp:
n/t


jmcarp:
Took snapshots, here's proof:

screen shot 2017-03-06 at 5 48 12 pm

PR #194, opened 2017-03-08
Add monitoring network outputs.

jmcarp:
@LinuxBozo


LinuxBozo:
Might as well add the AZs for each subnet as well?

jmcarp:
Agree, will update tonight.

jmcarp:
Updated, goes with https://github.com/18F/cg-deploy-prometheus/pull/1.

PR #196, opened 2017-03-14
Consolidate staging and production stacks.

jmcarp:
* Deploy staging and production from the same stack, with differences factored into variables
* Use `count` to skip unwanted ELB on staging

I'm hoping to get this ready before starting https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-3593 so that we don't copy and paste this stack again.


terraform/modules/client-elbs/star_18f_gov_elb.tf
```diff
@@ -1,4 +1,6 @@
 resource "aws_elb" "star_18f_gov_elb" {
+  count = "${var.count}"
```
LinuxBozo: Instead of using a `var` here, WDYT about using a conditional based on `stack_description`?
https://www.terraform.io/docs/configuration/interpolation.html#conditionals

terraform/stacks/main/variables.tf
```diff
@@ -33,25 +33,27 @@ variable "account_id" {}
...
 variable "services_cidr_2" {}
 variable "kubernetes_cluster_id" {}
 
 variable "restricted_ingress_web_cidrs" {}
 
-/* Variables for Diego Deployment */
+variable "client_elb_count" {
+  default = 0
+}
+variable "client_elb_cert_name" {
```
LinuxBozo: Any reason not to just name this exactly what it is? 18f_gov_elb? instead of client_elb?

terraform/modules/client-elbs/star_18f_gov_elb.tf
jmcarp: That's what I wanted to do originally, but we're still on terraform 0.7.13, and conditionals were added in 0.8.0. We're blocked on upgrading to latest by https://github.com/hashicorp/terraform/pull/12414, but we might be able to upgrade to a verison in the 0.8.x series now. Will check.

jmcarp:
The build will fail until https://github.com/18F/cg-deploy-concourse-docker-image/pull/18 is merged.

jmcarp:
BTW, props to @jseppi for catching this issue during pairing a few weeks back.

PR #197, opened 2017-03-14
Dev stack

jmcarp:
Builds on #196; most of the diffs come from that patch.

### DIFF
--- a/ci/credentials.yml.example
+++ b/ci/credentials.yml.example
@@ -31,6 +31,24 @@ concourse_staging_cidr: 10.0.8.0/24
 monitoring_production_cidr: CHANGEME
 monitoring_staging_cidr: CHANGEME
 
+development_vpc_cidr: 10.3.0.0/16
+development_public_cidr_1: 10.3.3.0/24
+development_public_cidr_2: 10.3.4.0/24
+development_private_cidr_1: 10.3.1.0/24
+development_private_cidr_2: 10.3.2.0/24
+development_rds_private_cidr_1: 10.3.5.0/24
+development_rds_private_cidr_2: 10.3.6.0/24
+development_services_cidr_1: 10.3.30.0/24
+development_services_cidr_2: 10.3.31.0/24
+development_kubernetes_cluster_id: kubernetes-staging
+development_rds_password: CHANGEME
+development_cf_rds_password: CHANGEME
+development_restricted_ingress_web_cidrs: "127.0.0.1/32,192.168.0.1/24"
+development_diego_cidr_1: 10.3.5.0/24
+development_diego_cidr_2: 10.3.6.0/24
+development_vpc_concourse_rds_password: CHANGEME
+development_vpc_concourse_cidr: 10.3.32.0/24
+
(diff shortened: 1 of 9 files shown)


jmcarp:
Note--this assumes that dev will use the same tls certs as staging, which we should change.

jmcarp:
Updated to restrict access to all dev elbs, h/t @cnelson

jmcarp:
@cnelson: wdyt about temporarily using the *.fr-stage cert for dev, so that we can continue with this story, and then pointing to the new dev cert later on when it's provisioned?

cnelson:
If it'll help with the terraform testing, SGTM but you won't actually be able to hit anything in a browser as the HSTS config for cloud.gov will prevent you from hitting a site with a CN mismatch.

jmcarp:
AFAICT, this story doesn't include anything we'd hit with the browser--looks like a dev stack in tf, dev bosh, and a dev jumpbox, plus sorting the staging / prod pipelines.

jmcarp:
Updated to temporarily use the staging certs for the dev stack to unblock other tasks in this story.

ci/credentials.yml.example
```diff
@@ -31,6 +31,24 @@ concourse_staging_cidr: 10.0.8.0/24
...
+development_vpc_cidr: 10.3.0.0/16
+development_public_cidr_1: 10.3.3.0/24
+development_public_cidr_2: 10.3.4.0/24
+development_private_cidr_1: 10.3.1.0/24
+development_private_cidr_2: 10.3.2.0/24
+development_rds_private_cidr_1: 10.3.5.0/24
+development_rds_private_cidr_2: 10.3.6.0/24
+development_services_cidr_1: 10.3.30.0/24
+development_services_cidr_2: 10.3.31.0/24
+development_kubernetes_cluster_id: kubernetes-staging
```
LinuxBozo: really a nit: this should probably be `kubernetes-development`

terraform/stacks/main/stack.tf
```diff
@@ -28,7 +28,9 @@ module "cf" {
     elb_main_cert_name = "${var.main_cert_name}"
     elb_apps_cert_name = "${var.apps_cert_name}"
     elb_subnets = "${module.stack.public_subnet_az1},${module.stack.public_subnet_az2}"
-    elb_security_groups = "${module.stack.web_traffic_security_group}"
+    elb_security_groups = "${var.force_restricted_network == "no" ?
```
LinuxBozo: 😍

terraform/stacks/main/stack.tf
```diff
@@ -56,6 +58,9 @@ module "diego" {
     stack_description = "${var.stack_description}"
     diego_cidr_1 = "${var.diego_cidr_1}"
     diego_cidr_2 = "${var.diego_cidr_2}"
+    ingress_cidrs = "${var.force_restricted_network == "no" ?
```
LinuxBozo: Huge 👍

LinuxBozo:
Love that the terraform upgrade is allowing for these conditionals. Merge when ready.

PR #198, opened 2017-03-16
Add static proxy ip.

sharms:


sharms:
This is a Forestry proof of concept where they are going to whitelist the ip in the DMZ to allow for secure communication

jmcarp:
Eventually a proxy server will get attached to this EIP.

terraform/stacks/tooling/outputs.tf
```diff
@@ -455,3 +455,8 @@ output "bosh_uaa_elb_dns_name" {
 output "bosh_uaa_elb_name" {
   value = "${aws_elb.bosh_uaa_elb.name}"
 }
+
+/* Static proxy */
+output "static_proxy_ip" {
```
rogeruiz: Do we want the name to be this generic? Is it possible that in the future we'd offer multiple EIPs? or would everyone just use the same one?

PR #203, opened 2017-03-24
Update bosh master username.

jmcarp:
Adds new bosh instances alongside existing instances with a non-default username. After backup and restore of the old instances, I'll send another patch to delete them.

### DIFF
--- a/terraform/modules/stack/base/outputs.tf
+++ b/terraform/modules/stack/base/outputs.tf
@@ -65,11 +65,20 @@ output "rds_postgres_security_group" {
 
 /* RDS Bosh Instance */
 output "bosh_rds_url" {
+  value = "${module.rds.rds_url}"
+}
+output "bosh_rds_96_url" {
   value = "${module.rds_96.rds_url}"
 }
 output "bosh_rds_host" {
-  value = "${module.rds_96.rds_host}"
+  value = "${module.rds.rds_host}"
 }
 output "bosh_rds_port" {
-  value = "${module.rds_96.rds_port}"
+  value = "${module.rds.rds_port}"
+}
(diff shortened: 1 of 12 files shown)


terraform/modules/stack/base/outputs.tf
```diff
@@ -65,11 +65,20 @@ output "rds_postgres_security_group" {
 
 /* RDS Bosh Instance */
 output "bosh_rds_url" {
```
LinuxBozo: Does it make sense to do the same pattern we've used elsewhere for output and suffix with current/previous?

jmcarp: IMO that works better for resources we can safely delete and recreate. For database instances, we'd need to `terraform state mv` any time we want to move a `_curr` instance to `_prev`.

LinuxBozo: Failing to see why given it's just the outputs? Why would we need to touch state?

jmcarp: So if we name the new instance `rds_curr`, at some point in the future we'll probably want to move it to `rds_prev` and put something else at `rds_curr`. Would terraform be smart enough to understand that rename, or would it delete and recreate the instance?

LinuxBozo: Again, it's not about the name of the resource itself, it's about the output..

So, we have the pattern currently
```
  output.foo_prev == resource.aws_something.foo_v1 
  output.foo_curr == resource.aws_something.foo_v2
```
Are you saying that terraform is going to destroy an instance based on the fact that the output reference changed?  That doesn't seem correct.

PR #206, opened 2017-03-30
[WIP] Vm instance profiles

jmcarp:
I still have more profiles to add, but could use thoughts on what I have so far.


terraform/modules/iam_role_policy/blobstore/policy.json
```diff
@@ -0,0 +1,13 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": ["s3:*"],
```
cnelson: image

Can you tighten up this action list to what's actually needed?

terraform/modules/iam_role_policy/cloudwatch/policy.json
```diff
@@ -0,0 +1,18 @@
...
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "logs:CreateLogGroup",
+        "logs:CreateLogStream",
+        "logs:PutLogEvents",
+        "logs:DescribeLogStreams",
+        "cloudwatch:ListMetrics",
```
cnelson: Do all machines need access to metrics to run aws logs?  The documented AWS policy is just:

```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "logs:CreateLogGroup",
        "logs:CreateLogStream",
        "logs:PutLogEvents",
        "logs:DescribeLogStreams"
    ],
      "Resource": [
        "arn:aws:logs:*:*:*"
    ]
  }
 ]
}
```

terraform/modules/iam_role_policy/blobstore/policy.json
jmcarp: Am I allowed to blame the bosh docs https://bosh.io/docs/aws-iam-instance-profiles.html ? Anyway, since this policy is meant to be part of the default role, can we restrict to `s3:GetObject`? I see that the blobstore bucket is a `public_encrypted_bucket`, so we could also drop this entirely.

terraform/modules/iam_role_policy/cloudwatch/policy.json
jmcarp: Looks like I added the metrics actions last year to use with https://github.com/riemann/riemann-tools/blob/master/tools/riemann-aws/bin/riemann-aws-rds-status. I was thinking we'd use that emitter on more deployments, but so far it's only running on concourse, and hasn't paged us since initial testing (so it's possibly broken or not useful). I'll drop these permissions here and add them to a separate policy that's attached to the atc instances.

cnelson: It looks like I also decided to depend on this being a thing in order to fail fast if the AWS creds are bad: https://github.com/18F/cg-awslogs-boshrelease/blame/master/jobs/awslogs/templates/bin/awslogsd.sh.erb#L23 but I guess we can drop that whole check (or find a better way to do it) once we aren't relying on creds.

jmcarp: Hm, WDYT about writing a separate story for logs / metrics permissions and handling it separately from creds -> profiles? Or is that what you're getting at?

cnelson: SGTM

PR #209, opened 2017-04-04
Drop unused roles and users.

jmcarp:
Need to double-check that none of these roles or users are still in use, and redeploy master bosh with instance profile.


terraform/stacks/tooling/outputs.tf
```diff
@@ -317,23 +316,6 @@ output "master_bosh_secret_access_key_curr" {
...
-output "tooling_bosh_secret_access_key_prev" {
-  value = "${module.tooling_bosh_user.secret_access_key_prev}"
-}
-output "tooling_bosh_access_key_id_curr" {
-  value = "${module.tooling_bosh_user.access_key_id_curr}"
-}
-output "tooling_bosh_secret_access_key_curr" {
-  value = "${module.tooling_bosh_user.secret_access_key_curr}"
-}
-
```
rogeruiz: So `cg-deploy-bosh` still uses this user for the following stanza. Looks like we'll need to bring it back. 😕

```yaml
- name: semver-tooling-version
  type: semver
  source:
    driver: s3
    bucket: {{semver-bucket}}
    key: {{semver-tooling-key}}
    access_key_id: {{semver-access-key-id-tooling}}
    secret_access_key: {{semver-secret-access-key-tooling}}
    region_name: {{aws-region}}
```

jmcarp: Sounds like we're using the wrong keys there.

rogeruiz: @jmcarp does it make sense to just create a new user for `semver`? one for master and tooling bosh? since it seems like we just need something to read/write to that particular bucket. wdyt?

PR #226, opened 2017-04-20
[WIP] Add vpc dns addresses.

jmcarp:
We're also outputting per-subnet dns addresses, but our bosh deployments seem to be using vpc-level dns addresses. WDYT @cnelson @LinuxBozo ?


jmcarp:
Goes with https://github.com/18F/cg-deploy-bosh/pull/73.

PR #228, opened 2017-04-25
Remove v1 of all IAM users

rogeruiz:
All these users have been rotated to `v2` / `_curr` values across all pipelines.

I will merge this and bootstrap the environments.

### DIFF
--- a/terraform/stacks/external/outputs.tf
+++ b/terraform/stacks/external/outputs.tf
@@ -3,10 +3,10 @@ output "cdn_broker_username" {
   value = "${module.cdn_broker.username}"
 }
 output "cdn_broker_access_key_id_prev" {
-  value = "${module.cdn_broker.access_key_id_prev}"
+  value = ""
 }
 output "cdn_broker_secret_access_key_prev" {
-  value = "${module.cdn_broker.secret_access_key_prev}"
+  value = ""
 }
 output "cdn_broker_access_key_id_curr" {
   value = "${module.cdn_broker.access_key_id_curr}"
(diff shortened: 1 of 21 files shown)


rogeruiz:
Gonna start before merging by deactivating all the previous IAM crews manually and monitor the CI. I've double-triple checked all the crews currently flown, but I want to make absolutely sure 🤞

jmcarp:
Looks like this doesn't actually delete credentials--it just hides them from terraform outputs. This should also finish the cdn broker credentials rotation.

rogeruiz:
Alright let me verify all the concourse creds are right and double check EW.

rogeruiz:
Thanks @jmcarp I'll remove the versions as well. Marked this WIP until for now.

- [ ] verify EW
- [ ] actually delete `_v2` keys in Terraform files.

wjwoodson:
Deleted all _v1 keys from Terraform files @rogeruiz and I are still in the process of verifying they're cleared from pipelines

rogeruiz:
Verified that the users are no longer in use in fly or manifests

terraform/stacks/external/outputs.tf
```diff
@@ -3,10 +3,10 @@ output "cdn_broker_username" {
   value = "${module.cdn_broker.username}"
 }
 output "cdn_broker_access_key_id_prev" {
-  value = "${module.cdn_broker.access_key_id_prev}"
+  value = ""
```
jmcarp: Why change this? Looks like the output from the module is already blank.

rogeruiz: ah okay. so we don't need to do it both places. let me revert this then.

rogeruiz: so does the same thing revert need to happen for `terraform/modules/cloudfoundry/outputs.tf` and `terraform/stacks/main/outputs.tf` as well? @jmcarp

jmcarp: That's the idea--we should only have to blank out each credential pair in one place.

PR #230, opened 2017-04-28
Make sure we're creating a default db for the RDS db user

LinuxBozo:


jmcarp:
Might be useful to add a comment about the issue we're fixing so we (I) remember what this is for later.

PR #231, opened 2017-04-28
default db fixes

LinuxBozo:
turns out printf quoting bits are bash specific.
Also, need to use awk to filter the database list to just the first column, since the user will match in the table for every other created database, so won't actually do the create.


LinuxBozo:
Also, added the note about the reasoning for creating the default db as noted in previous PR #230

PR #237, opened 2017-05-09
Manage billing buckets and permissions.

jmcarp:
cc @jcscottiii


cnelson:
Looks like the billing_user definition is missing if I'm reading that PR check right

jmcarp:
Forgot to check it in 😓

PR #240, opened 2017-05-11
Add instance profile for tagged concourse workers.

jmcarp:


jmcarp:
TODO: can we lock down the policy more than not at all? cc @cnelson @LinuxBozo

jmcarp:
This needs another look now that we're managing admin permissions for operators; wdyt @cnelson ?

terraform/stacks/tooling/stack.tf
```diff
@@ -313,3 +320,18 @@ resource "aws_iam_policy_attachment" "concourse_worker" {
...
   ]
 }
+
+module "operators" {
+  source = "../../modules/operators"
+}
+
+resource "aws_iam_policy_attachment" "admin" {
+  name = "admin"
+  policy_arn = "arn:aws-us-gov:iam::aws:policy/AdministratorAccess"
```
cnelson: This is ok to be here twice (other in external stack) because external runs in the other non-govcloud aws account right?

terraform/stacks/external/stack.tf
```diff
@@ -13,3 +13,19 @@ module "limit_check_user" {
...
+}
+
+resource "aws_iam_policy_attachment" "admin" {
+  name = "admin"
+  count = "${var.stack_description == "production" ? 1 : 0}"
+  policy_arn = "arn:aws:iam::aws:policy/AdministratorAccess"
+  roles = [
+    "${concat(
+      module.operators.operators,
+      list("terraform-provision")
```
cnelson: I have no idea if cross-partition account delegation works, but should we give that a go rather than keeping this AWS cred around? Or we can punt it for now and make a card to try it out in a future sprint...

terraform/stacks/tooling/stack.tf
jmcarp: Yep, and the policy arn is slightly different (aws-us-gov vs aws).

LinuxBozo: We're probably going to have to make further note about standardizing operator account names across accounts as well to make this all work.

terraform/stacks/external/stack.tf
jmcarp: We should definitely investigate this, but in the interest of getting this story done, I vote for a separate card.

cnelson: SGTM

terraform/modules/iam_role_policy/concourse_iaas_worker/policy.json
```diff
@@ -0,0 +1,10 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
```
cnelson: I want to add a comment here to state this is intended to be a clone of arn:aws-us-gov:iam::aws:policy/AdministratorAccess so future people don't make the same mistake we did / wonder why we have such a broad policy in place, but no comments in json.

WDYT about a comment in the TF where this is referenced?

PR #265, opened 2017-06-02
Attach ElasticIP to NAT gateway instances in each VPC

LinuxBozo:
Will more than likely need to schedule maintenance window to put this in place since it should recreate the NAT instances with latest AMI image.

### DIFF
--- a/terraform/modules/bosh_vpc/network_private.tf
+++ b/terraform/modules/bosh_vpc/network_private.tf
@@ -101,7 +101,7 @@ resource "aws_instance" "az1_private_nat" {
   ami = "ami-e8ab1489"
   instance_type = "${var.nat_gateway_instance_type}"
   source_dest_check = false
-
+  associate_public_ip_address = true
   subnet_id = "${aws_subnet.az1_public.id}"
 
   vpc_security_group_ids = ["${aws_security_group.local_vpc_traffic.id}"]
@@ -114,7 +114,7 @@ resource "aws_instance" "az2_private_nat" {
   ami = "ami-e8ab1489"
   instance_type = "${var.nat_gateway_instance_type}"
   source_dest_check = false
-
+  associate_public_ip_address = true
   subnet_id = "${aws_subnet.az2_public.id}"
 
   vpc_security_group_ids = ["${aws_security_group.local_vpc_traffic.id}"]
--- a/terraform/modules/bosh_vpc/network_private.tf
+++ b/terraform/modules/bosh_vpc/network_private.tf
@@ -75,7 +75,6 @@ resource "aws_instance" "az1_private_nat" {
   ami = "${var.nat_gateway_ami}"
   instance_type = "${var.nat_gateway_instance_type}"
   source_dest_check = false
-  associate_public_ip_address = true
 
   subnet_id = "${aws_subnet.az1_public.id}"
 
@@ -90,7 +89,6 @@ resource "aws_instance" "az2_private_nat" {
   ami = "${var.nat_gateway_ami}"
   instance_type = "${var.nat_gateway_instance_type}"
   source_dest_check = false
-  associate_public_ip_address = true
 
   subnet_id = "${aws_subnet.az2_public.id}"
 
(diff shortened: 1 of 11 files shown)


cnelson:
Could we spin up new nat gateways with EIPs + alter the route tables to use those, then in another TF update drop the old nat gateways for zero downtime rollout?

terraform/modules/bosh_vpc/network_private.tf
```diff
@@ -100,3 +98,13 @@ resource "aws_instance" "az2_private_nat" {
     Name = "${var.stack_description}  (AZ2 NAT)"
   }
 }
+
+resource "aws_eip" "az1_nat_eip" {
```
cnelson: Should probably add  `lifecycle {prevent_destroy = true}` to these resources, as these are going to end up in other people's firewall configurations and if one gets deleted on accident, there's no getting that same ip back from AWS again

LinuxBozo:
@cnelson Good call. The rotation strategy tis probably something we should do anyway for future AMI updates.

terraform/modules/bosh_vpc/variables.tf
```diff
@@ -41,7 +41,7 @@ variable "nat_gateway_instance_type" {
 }
 
 variable "nat_gateway_ami" {
-  default = "ami-e8ab1489"
+  default = "ami-6ae2660b"
```
jmcarp: Would it make sense to get the AMI ID programmatically instead of hard-coding, possibly using https://www.terraform.io/docs/providers/aws/d/ami.html? If not, maybe we can add a comment explaining what this refers to--or we can just wait for govcloud to offer managed nat gateways :trollface:

terraform/modules/bosh_vpc/network_private.tf
jmcarp: This is both unrelated and unimportant, but I have style questions:
* Can we standardize on where `az1` goes in resource names--prefix or suffix?
* Same question for including resource types in resource names--since resources are namespaced by type, we could call this one `nat_az1` without ambiguity--but maybe it's clearer to keep the type in the name.

terraform/modules/bosh_vpc/variables.tf
cnelson: Given the convo above about it being a multi step process to zero downtime upgrade these instances I don't think we'd want to autobump this.

terraform/modules/bosh_vpc/network_private.tf
```diff
@@ -99,12 +124,29 @@ resource "aws_instance" "az2_private_nat" {
   }
 }
 
-resource "aws_eip" "az1_nat_eip" {
-  instance = "${aws_instance.az1_private_nat.id}"
-  vpc = true
+resource "aws_instance" "az1_private_nat_2017_03" {
```
cnelson: What's 2017-03?  Is that the date on the AMI?

cnelson: What's 2017-03?  Is that the date on the AMI?

terraform/modules/bosh_vpc/network_private.tf
```diff
@@ -48,31 +47,58 @@ resource "aws_route_table" "az2_private_route_table" {
...
 }
 
+resource "aws_route_table_association" "az1_private_rta" {
+  subnet_id = "${aws_subnet.az1_private.id}"
+  route_table_id = "${aws_route_table.az1_private_route_table.id}"
+}
+
+resource "aws_route_table_association" "az2_private_rta" {
+  subnet_id = "${aws_subnet.az2_private.id}"
+  route_table_id = "${aws_route_table.az2_private_route_table.id}"
```
cnelson: When we roll this out to staging we should verify that there's no disruption in service doing the route table change in the same TF update as creating the new Gateways.  I'm a little worried that the route table change will be applied before the new VMs are fully booted and forwarding packets.

LinuxBozo: The first time this is run, it's still using the older NAT instances, so shouldn't be any disruption. Once the newer instances are created, then we can go back and update the routes. This is why I documented this bit in the comments about how to rotate. New instances declared first, then update routes and EIP instances.

terraform/modules/bosh_vpc/network_private.tf
LinuxBozo: Yep.

terraform/modules/bosh_vpc/network_private.tf
cnelson: I mis-read this, LGTM;

PR #275, opened 2017-06-13
Removing AWS creds

rogeruiz:
Sacrificing :lock_with_ink_pen: to the :volcano:

Pipeline still needs to fly.

6e3a153d-e8d3-4914-8236-d79205204dfa_text_hi


rogeruiz:
I'm confident that it'll fly fine, but didn't want to do it on a monday night 🦉

pburkholder:
LGTM. Running `plan` should be sufficient to assure this works with roles, right?

ci/pipeline.yml
```diff
@@ -146,15 +146,15 @@ jobs:
     - get: plan-timer
       trigger: true
   - task: plan-update-tooling
+    tags:
```
jmcarp: Do we need to tag the apply steps too?

rogeruiz: ah yeah. thanks! 😆

rogeruiz: @pburkholder let's pair on these changes that we need to do after standup

cnelson:
I wouldn't nuke the stemcell user until we have resolution from AWS on the `ImportVolume` bug/issue.

rogeruiz:
ah true. that was my mistake for mentioning that one to @pburkholder. _in a perfect world right?_ :trollface: I'll revert just those changes

PR #276, opened 2017-06-13
Include cdn broker iam path

pburkholder:
Reviewing #275 I noticed the example creds were missing these.

### DIFF
--- a/ci/credentials.yml.example
+++ b/ci/credentials.yml.example
@@ -95,13 +95,13 @@ aws_external_s3_tfstate_bucket: EXTERNAL_TERRAFORM_BUCKET
 cdn_broker_username_staging: cdn-broker-staging
 cdn_broker_bucket_staging: cdn-broker-le-verify-staging
 cdn_broker_cloudfront_prefix_staging: cg-staging/*
-cdn_broker_hosted_zone_staging: ZZZZZZZZZZZZZZ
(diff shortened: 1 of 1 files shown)


ci/credentials.yml.example
```diff
@@ -95,12 +95,14 @@ aws_external_s3_tfstate_bucket: EXTERNAL_TERRAFORM_BUCKET
 cdn_broker_username_staging: cdn-broker-staging
 cdn_broker_bucket_staging: cdn-broker-le-verify-staging
 cdn_broker_cloudfront_prefix_staging: cg-staging/*
-cdn_broker_hosted_zone_staging:
+cdn_broker_hosted_zone_staging: ZZZZZZZZZZZZZZ
```
jmcarp: Is `ZZZZZZZZZZZZZZ` more descriptive than the empty string?

pburkholder: Probably not, so changed to ZONEIDSTRING01, which might be.

jmcarp: As an aside, this is why I think example credentials are rarely valuable--our variable names should be descriptive enough in most cases, and if they're not, better to rename than add examples.

pburkholder: I have nothing vested in the color of this 🚲  🏚 , but I thought it marginally useful to hint that the value would be a zone id starting with 'Z'. 

Want to merge as is? Or do you prefer an empty string?

PR #278, opened 2017-06-19
Add mssql and oracle security groups.

jmcarp:
Part of https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-5899.


pburkholder:
Should https://github.com/18F/cg-provision/blob/rds-enterprise/terraform/modules/stack/base/base.tf#L48 also reference the new SGs?

Tangentially, why does https://github.com/18F/cg-provision/blob/rds-enterprise/terraform/stacks/main/stack.tf#L139 include both the SGs for postgres and mysql? Are they both needed by Concourse?  (ditto for https://github.com/18F/cg-provision/blob/rds-enterprise/terraform/stacks/tooling/stack.tf#L38 and https://github.com/18F/cg-provision/blob/rds-enterprise/terraform/stacks/tooling/stack.tf#L57)

jmcarp:
@pburkholder: no, the first line you linked to creates the bosh rds instance, which uses postgres. Same deal for concourse--we always use postgres, so we can drop the other rds security groups. I'll update this today to drop those unnecessary sgs.

PR #281, opened 2017-06-23
drop most users, add new access keys for the remaining users

cnelson:
Dropping all users except for:
- cdn_broker
- stemcell
- limit_checker (external)
- billing_user (customer squad)


terraform/stacks/tooling/outputs.tf
```diff
@@ -409,40 +375,6 @@ output "stemcell_secret_access_key_curr" {
   value = "${module.stemcell_user.secret_access_key_curr}"
 }
 
-/* limit check user */
-output "limit_check_username" {
```
LinuxBozo: Looks like limit check user is still in use? Don't you still need these outputs?

terraform/stacks/tooling/stack.tf
```diff
@@ -116,41 +116,13 @@ module "billing_user" {
...
-}
-
 module "stemcell_user" {
   source = "../../modules/iam_user/stemcell_user"
   username = "stemcell"
   aws_partition = "${var.aws_partition}"
   stemcell_bucket = "cg-stemcell-images"
 }
 
-module "limit_check_user" {
```
LinuxBozo: Again, don't you still need limit check user?

PR #284, opened 2017-06-23
Yank the influxdb backup related bits

LinuxBozo:
Since we've yanked out the influxdb backups, we should be able to yank all this now, yeah?


cnelson:
We should merge 

https://github.com/18F/cg-deploy-monitoring/pull/136

and 

https://github.com/18F/cg-deploy-bosh/pull/112 

before merging this so we don't have any dependencies on it existing

PR #287, opened 2017-06-23
Help for local planning

pburkholder:


pburkholder:
I don't like this this is just a README, but it seems to be potentially useful then formalize into something executable.  @wjwoodson @rogeruiz

PR #294, opened 2017-07-26
Add internal logsearch load balancer.

jmcarp:
So that external tasks and jobs, like cg-billing and cg-deploy-kubernetes, can find elastic masters by load balancer instead of using static ips.

### DIFF
--- a/terraform/modules/kubernetes/elb.tf
+++ b/terraform/modules/kubernetes/elb.tf
@@ -20,7 +20,7 @@ resource "aws_elb" "kubernetes_elb" {
     interval = 30
   }
 
-  tags = {
+  tags {
     Name = "${var.stack_description}-kubernetes"
   }
 }
--- a/terraform/modules/kubernetes/sg_ec2.tf
+++ b/terraform/modules/kubernetes/sg_ec2.tf
@@ -2,7 +2,7 @@ resource "aws_security_group" "kubernetes_ec2" {
   description = "Allow access to incoming kubernetes traffic"
   vpc_id = "${var.vpc_id}"
 
-  tags = {
+  tags {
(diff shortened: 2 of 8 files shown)


jmcarp:
More specifically, I think this gets us closer to being able to drop static ips from the masters. The services network that we're using for logsearch uses type `manual`, so bosh links use ip addresses instead of dns names in templates. That means if we recreate the masters, the data nodes won't be able to find the masters until the data nodes redeploy. But I'm not sure the 0th data node will ever pass its `post-start` check, since the other data nodes won't be in the cluster, and it will never turn green. I haven't tested this yet, so maybe elastic will be smarter than I'm guessing. Anyway, I think the right solution is to either change the the network type to `dynamic` so that bosh uses dns names in templates--which is blocked until kubernetes switches to dynamic addresses too, since we're using the same network in cloud config--or tell bosh to always use dns names with https://github.com/cloudfoundry/bosh/blob/master/jobs/director/spec#L144-L146.

cnelson:
I'm sure it would take some upstream work, but WDYT about adding support for https://www.elastic.co/guide/en/elasticsearch/plugins/5.5/discovery-ec2-discovery.html and finding masters via tags?

jmcarp:
I like the idea of not depending on bosh dns or static ips for service discovery. Seems like this would work better if bosh supported tags per instance group so that we don't have to do it with a `pre-start` script like we do for kubernetes.

cnelson:
I think we could do this without using pre-start, by adding a new tag to identify the environment dev/stage/prod/etc and then use that + the exiting job tag bosh adds to find the masters.

Something like:
```yaml
discovery.ec2.tag.environment: dev
discovery.ec2.tag.job: elasticsearch_master
```

PR #297, opened 2017-08-14
[WIP] Add platform logsearch load balancers.

jmcarp:
* Add platform logs kibana ELB for restricted network access
* Add platform logs syslog ELB (internal) to load-balance ingestors without running haproxy

### DIFF
--- a/terraform/modules/logsearch/elb_platform_kibana.tf
+++ b/terraform/modules/logsearch/elb_platform_kibana.tf
@@ -1,0 +1,26 @@
+resource "aws_elb" "platform_kibana_elb" {
+  name = "${var.stack_description}-platform-kibana"
+  subnets = ["${var.elb_subnets}"]
+  security_groups = ["${var.restricted_security_group}"]
+  idle_timeout = 3600
+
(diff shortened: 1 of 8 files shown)


jmcarp:
Yep, let's not merge until platform logs are working on dev.

terraform/modules/logsearch/elb_platform_kibana.tf
```diff
@@ -0,0 +1,25 @@
+resource "aws_elb" "platform_kibana_elb" {
+  name = "${var.stack_description}-platform-kibana"
+  subnets = ["${var.elb_subnets}"]
+  security_groups = ["${var.restricted_security_group}"]
+  idle_timeout = 3600
+
+  listener {
+    instance_port = 80
```
LinuxBozo: I get this is in the restricted web traffic security group, but still should be HTTPS with `*` cert, not plain HTTP, yeah?

jmcarp: Updated, WDYT?

PR #303, opened 2017-09-15
Add cg_semver_user for read-only bucket access

jcscottiii:
@rogeruiz


rogeruiz:
Premature merge here. whoops.

PR #318, opened 2017-10-27
tooling bosh_static_ip

wjwoodson:


terraform/stacks/tooling/outputs.tf
```diff
@@ -107,6 +107,9 @@ output "master_bosh_static_ip" {
 output "tooling_bosh_static_ip" {
   value = "${cidrhost("${var.private_cidr_1}", 7)}"
 }
+output "bosh_static_ip" {
+  value = "${cidrhost("${var.private_cidr_1}", 7)}"
+}
```
pburkholder: Why is `bosh_static_ip` needed here when `tooling_bosh_static_ip` also uses `cidrhost("${var.private_cidr_1}", 7)`?

wjwoodson: This is specifically to enable secret generation for all bosh envs in https://github.com/18F/cg-secret-rotation/blob/master/ci/update-certificates.sh#L32, but also to standardize our tf outputs to include both `bosh_static_ip` for local bosh, and `tooling_bosh_static_ip` for shared use (DNS resolution)

PR #320, opened 2017-11-19
use the VPC flow log Terraform module

afeld:
This reusable module was built as part of the DevSecOps effort, and we are looking for adoption and feedback. Rather than every project needing to include the same boilerplate code (for setting up the flow logs, in this case), hoping to get various projects at GSA to use reusable pieces like this one. Let me know what you think!

I `valididate`d the module locally, but don't have the ability to test it for real. Since the `aws_cloudwatch_log_group` changed modules, you will need to `terraform state mv` it to ensure the log group isn't destroyed and recreated.

### DIFF
--- a/terraform/modules/bosh_vpc/vpc.tf
+++ b/terraform/modules/bosh_vpc/vpc.tf
@@ -15,57 +15,10 @@ resource "aws_vpc" "main_vpc" {
 
 }
 
-# Create CloudWatch log group
-resource "aws_cloudwatch_log_group" "main_vpc_flow_log_cloudwatch_log_group" {
-  name = "${var.stack_description}-flow-log-group"
-}
+module "flow_logs" {
+  source = "github.com/GSA/terraform-vpc-flow-log"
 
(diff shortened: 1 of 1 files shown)


jmcarp:
Thanks for letting us know about your module! Still, I'd rather not import it from your repo. Like you said, that would require editing our state files, which I don't want to do. And it would mean you wouldn't be able to change that module without potentially requiring us to make more state changes, or otherwise breaking things on our side. Unless you want to start versioning your modules and promise to follow semver, which sounds like more trouble than it's worth.

afeld:
> start versioning your modules and promise to follow semver

I have been: https://github.com/GSA/terraform-vpc-flow-log/releases. Happy to switch to that if you prefer.

> sounds like more trouble than it's worth

This is a small first step, but the idea is that we'd like to have more modules that build up to each system not having to maintain the code on their own. The plan is to write up canonical security control narratives and have each component assessed on its own, so that they can build up to take more and more compliance responsibility off systems like cloud.gov.

PR #324, opened 2017-12-09
wait for propagation of the instance profile

afeld:
It's a bit of a hack, but this will help with the gap between when the Instance Profile is created, and when it's actually usable. In leveraging this module, trying to use the Instance Profile immediately was giving an error of `Invalid IAM Instance Profile name`.

/cc @Vermyndax


jmcarp:
Looks like this was resolved in https://github.com/terraform-providers/terraform-provider-aws/pull/3055.

PR #328, opened 2017-12-24
Use built-in list types instead of splitting strings.

jmcarp:
Now that HCL mostly supports list variables, we don't have to split strings anymore. I pointed our pipeline at this branch and verified that nothing will change in any of our environments.

### DIFF
--- a/terraform/modules/bosh_vpc/sg_restricted_web_traffic.tf
+++ b/terraform/modules/bosh_vpc/sg_restricted_web_traffic.tf
@@ -14,14 +14,14 @@ resource "aws_security_group" "restricted_web_traffic" {
     from_port = 80
     to_port = 80
     protocol = "tcp"
-    cidr_blocks = ["${split(",", var.restricted_ingress_web_cidrs)}"]
+    cidr_blocks = ["${var.restricted_ingress_web_cidrs}"]
   }
 
   ingress {
     from_port = 443
     to_port = 443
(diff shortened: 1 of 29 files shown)


terraform/modules/bosh_vpc/sg_restricted_web_traffic.tf
```diff
@@ -14,14 +14,14 @@ resource "aws_security_group" "restricted_web_traffic" {
     from_port = 80
     to_port = 80
     protocol = "tcp"
-    cidr_blocks = ["${split(",", var.restricted_ingress_web_cidrs)}"]
+    cidr_blocks = ["${var.restricted_ingress_web_cidrs}"]
```
LinuxBozo: Somehow, this doesn't quite make sense to me. If this supports passing in a list, why is the surround `[ ]` still necessary? To me, this now reads incorrectly and is less understandable.

jmcarp: Agreed that we shouldn't have to wrap lists in square brackets. There's an issue for this at https://github.com/hashicorp/terraform/issues/13103, and hopefully it'll get fixed eventually. Still, I'd rather pass around lists than delimited strings, and the extra brackets problem isn't new--we had to wrap the `${split(...)}` in brackets too, so we're using pointless brackets either way.

PR #329, opened 2017-12-26
Drop riemann and static ips.

jmcarp:
* Drop unused riemann grafana elb (we have a separate elb for prometheus)
* Drop unused policy and static ips

### DIFF
--- a/terraform/modules/monitoring/sg_monitoring.tf
+++ b/terraform/modules/monitoring/sg_monitoring.tf
@@ -8,44 +8,34 @@ resource "aws_security_group" "monitoring" {
   description = "Allow access to incoming monitoring traffic"
   vpc_id = "${var.vpc_id}"
 
-  tags =  {
+  tags {
     Name = "${var.stack_description} - Incoming Monitoring Traffic"
   }
-
 }
 
 resource "aws_security_group_rule" "self_reference" {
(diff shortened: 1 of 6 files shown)


terraform/modules/monitoring/sg_monitoring.tf
```diff
@@ -8,44 +8,34 @@ resource "aws_security_group" "monitoring" {
...
 }
 
-resource "aws_security_group_rule" "monitoring_riemann" {
-    type = "ingress"
-    from_port = 5555
-    to_port = 5559
-    protocol = "all"
-    cidr_blocks = ["0.0.0.0/0"]
-    security_group_id = "${aws_security_group.monitoring.id}"
+resource "aws_security_group_rule" "monitoring_nginx" {
```
LinuxBozo: Looks like this rule is being added, but not used. What is it for?

jmcarp: This allows access from the load balancer to nginx on 8080. It's redundant with https://github.com/18F/cg-provision/blob/master/terraform/modules/bosh_vpc/sg_bosh.tf#L72-L79, but eventually I think we should drop those overgeneral rules from sg-bosh and move them into security groups that are specific to a deployment.

PR #337, opened 2018-01-08
Flatten *.18f.gov elb name to simplify merge.

jmcarp:
.........................


jmcarp:
https://ih0.redbubble.net/image.149471843.4713/flat,1000x1000,075,f.jpg

PR #387, opened 2018-03-28
[WIP] Remove dev pipeline

timothy-spencer:
This change is to try to remove the development pipeline/environment from everything so that we can just have one way of deploying a staging/prod setup into multiple AWS accounts.  This is related to https://github.com/18F/cg-deploy-bosh/pull/245


scripts/bootstrap/07-deploy-concourse.sh
```diff
@@ -5,14 +5,15 @@ set -eux
...
+  --config ../cg-deploy-concourse/ci/pipeline.yml \
   --load-vars-from ${WORKSPACE_DIR}/concourse-environment.yml \
   --load-vars-from ../cg-deploy-concourse/ci/concourse-defaults.yml \
   --var secrets-bucket=${VARZ_BUCKET} \
   --var concourse-production-deployment-bosh-target=$(bosh interpolate ${WORKSPACE_DIR}/tooling-state.yml --path /terraform_outputs/tooling_bosh_static_ip) \
   --var concourse-production-private-passphrase=${CONCOURSE_SECRETS_PASSPHRASE} \
   --var tf-state-bucket=${TF_STATE_BUCKET} \
-  --var slack-webhook-url=${SLACK_WEBHOOK_URL}
+  --var slack-webhook-url=${SLACK_WEBHOOK_URL} \
+  --ops-file bosh/opsfiles/development.yml
```
jmcarp: Does this work? I tried passing `--ops-file` to `fly set-pipeline` and got ```error: unknown flag `ops-file'```.

timothy-spencer: Hmm.  You are right.  I will get a bosh int in there.  I haven't worked myself back to where I can use this yet.  I plan on rebuilding the whole env from scratch, but wanted to work through the issues that I could work through with the existing setup first.

Thanks!

ci/credentials.yml.example
```diff
@@ -8,13 +10,25 @@ cg_provision_development_git_branch: master
...
+# This is where the environment is deployed (AWS govcloud)
 aws_access_key_id: CHANGEME
 aws_secret_access_key: CHANGEME
 aws_default_region: CHANGEME
 aws_s3_tfstate_bucket: CHANGEME
+aws_s3_cloudtrail_bucket: CHANGEME
 aws_az1: CHANGEME
 aws_az2: CHANGEME
 
+lets_encrypt_hosted_zone_staging: CHANGEME
```
jmcarp: Very minor, but I don't really see how these examples help. When most of the values are CHANGEME, this file isn't much more than a list of the variables we use in the pipeline, which we can already get from the pipeline. I'm not asking for a change here, just wondering if there's value in maintaining these example files.

timothy-spencer: I think the idea is that this is for when you don't have a pipeline to start with, like when you are bootstrapping everything.  I've also used this when grepping as a sign that "oh, this is in the pipeline, I'll go look there".

bosh/opsfiles/development.yml
```diff
@@ -0,0 +1,524 @@
...
+# we have to edit all sections.  Sigh.
+
+# these are taken care of in prod.
+- type: remove
+  path: /jobs/name=plan-dns?
+
+- type: remove
+  path: /jobs/name=apply-dns?
+
+## plan-bootstrap-tooling
```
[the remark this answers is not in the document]
timothy-spencer: Yeah, that whole approach got a bit... larger than expected, I think.  I did it this way because Will said that this was a good approach.  If we have one pipeline that we alter according to the environment it is being deployed into, that's not unreasonable.  I think that we both assumed that the number of changes would be small, and thus this would end up being a much smaller change to the overall project than me refactoring all this stuff to be variables.  It did keep the changes to the project in other areas fairly minimal, but it turns out that the opsfile ended up getting big and hairy.

I'm not sure where to go with this.  I think that it's not a bad approach.  I like having only one pipeline to maintain.  But we certainly could shrink this opsfile a lot if we moved a lot of these things into variables.  Do you see that as a clear win over just keeping these changes encapsulated in this opsfile?

sharms:
Hi folks,

Excellent PR - it got a little large, so we reviewed each section and detailed next steps.  I am going to close this one and we will open smaller PRs to move forward with the following changes:
* Update documentation to remove steps with 'seems' and references to race conditions.  These conditions should be solved as part of the process
* Eliminate BOSH blobstore copies - builds should be reproducable and bootstrapped as part of BOSH releases pipeline
* Remove fly intercept documentation for ALB certs - these secrets can be derived without intercepting live containers
* Accept using standard pipeline over development (line 87 in README.md)
* Remove bootstrap opsfile (bootstrap will use bucc)
* Accept partial development opsfile and instructurions to run pipeline through BOSH int.  
  - Remove interpolation of variables which concourse does natively and migrate to concourse-defaults.yml structure
* Accept bosh/opsfiles/production.yml removal of IAAS tags from bootstrap instance
* Remove tspencer/removedevpipeline branch hard coding
* Reject aws_s3_tfstate_bucket to cg-s3-cloudtrail - tf state is seperate from logs
* Accept Slack message including environment, but change it to not require 'meta'
* Accept TF Var cloudtrail bucket
* Remove redundant sed for 'iaas' tags, should be opsfile 
* Accept using variables for cloudtrail buckets

PR #390, opened 2018-04-02
[WIP] Drop legacy ELBs.

jmcarp:
Don't merge until https://github.com/18F/cg-deploy-bosh/pull/249 is rolled out.


sharms:
Prior to merging verified that the list of ELBS with these suffixes have no instances attached, rendering them currently out of service:
* 18f-gov
* CloudFoundry-Apps
* CloudFoundry-Logging
* CloudFoundry-Main
* Concourse-$az
* prometheus
* shibboleth-proxy
* bosh-uaa
* nessus

PR #392, opened 2018-04-04
Add ipv6 support to vpcs, subnets, and lbs.

jmcarp:
Restore ipv6 🎸

### DIFF
--- a/terraform/modules/bosh_vpc/network_public.tf
+++ b/terraform/modules/bosh_vpc/network_public.tf
@@ -13,6 +13,8 @@
 resource "aws_subnet" "az1_public" {
   vpc_id = "${aws_vpc.main_vpc.id}"
   cidr_block = "${var.public_cidr_1}"
+  // Hack: Default to a valid ipv6 cidr to handle empty vpc cidr on initial plan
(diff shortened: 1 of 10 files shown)


terraform/modules/bosh_vpc/network_public.tf
```diff
@@ -13,6 +13,8 @@
 resource "aws_subnet" "az1_public" {
   vpc_id = "${aws_vpc.main_vpc.id}"
   cidr_block = "${var.public_cidr_1}"
+  // Hack: Default to a valid ipv6 cidr to handle case when vpc cidr doesn't exist yet
+  ipv6_cidr_block = "${cidrsubnet(aws_vpc.main_vpc.ipv6_cidr_block != "" ? aws_vpc.main_vpc.ipv6_cidr_block : "fd00::/8", 8, 0)}"
```
LinuxBozo: Is there a time where it doesn't exist yet? Any way we can use a `depends_on` reference to the vpc to force it to exist before making the change here?

jmcarp: This happens on the first `terraform plan`. The main vpc ipv6 cidr block hasn't been created yet, so its value is `""`, and calling `cidrsubnet` with an empty string is an error. This seems like a bug/limitation in terraform, but I haven't found a less bad workaround. Updated the comments.

PR #394, opened 2018-04-05
Add static egress to tooling nat gateways.

jmcarp:
So that we can configure ip-based auth for gsa smtp.


terraform/stacks/tooling/variables.tf
```diff
@@ -147,3 +147,7 @@ variable "billing_bucket" {
 variable "cg_binaries_bucket" {
    default = "cg-binaries"
 }
+
+variable "use_nat_gateway_eip" {
+  default = false
```
LinuxBozo: any reason not to just default this to true and yank the override in the pipeline?

PR #424, opened 2018-05-15
(WIP) - Provision cidr blocks for cf services using terraform

jontours:
I _did_ verify that the inputs to the function are correct by using `ipcalc`.  However I'm not confident that this is all wired together correctly.


terraform/stacks/main/outputs.tf
```diff
@@ -136,7 +136,7 @@ output "logsearch_static_ips" {
 data "template_file" "kubernetes_static_ips" {
   count = 31
   vars {
-    address = "${cidrhost("${var.services_cidr_1}", "${count.index + 223}")}"
+    address = "${cidrhost("${module.services_cidr_1}", "${count.index + 223}")}"
```
jmcarp: Terraform validation is choking here:

```
Error: Error loading /tmp/build/49f2779f/pull-request/terraform/stacks/main/outputs.tf: Error reading config for template_file[kubernetes_static_ips]: module.services_cidr_1: module variables must be three parts: module.name.attr in:
```

terraform/modules/cloudfoundry/variables.tf
```diff
@@ -63,14 +63,10 @@ variable "vpc_id" {}
 variable "private_route_table_az1" {}
 
 variable "private_route_table_az2" {}
+# Assuming these are now being defined by terraform
```
jmcarp: I think this information is already covered by the fact that the cidr variables don't have defaults 😄

terraform/stacks/main/outputs.tf
```diff
@@ -103,30 +103,30 @@ output "services_subnet_az2" {
   value = "${module.cf.services_subnet_az2}"
 }
 output "services_subnet_cidr_az1" {
-  value = "${var.services_cidr_1}"
+  value = "${module.cf.services_cidr_1}"
```
jmcarp: Looks like the `cloudfoundry` module doesn't define an output with this name--we'll need to add one.

terraform/stacks/main/outputs.tf
```diff
@@ -103,30 +103,30 @@ output "services_subnet_az2" {
...
 }
 output "services_subnet_cidr_az2" {
-  value = "${var.services_cidr_2}"
+  value = "${module.cf.services_cidr_2}"
 }
 output "services_subnet_gateway_az1" {
-  value = "${cidrhost("${var.services_cidr_1}", 1)}"
+  value = "${cidrhost("${module.cf.services_cidr_1}", 1)}"
```
jmcarp: I see that this issue predates your patch, but I don't think we need all the quotes and curly braces here--I think we can use

```
value = "${cidrhost(module.cf.services_cidr_1, 1)}"
```

jmcarp:
Looks good! I left a few comments on the diffs--give me or @sharms a call and we can resolve and merge. BTW, you can run the `validate.sh` script inside this repo to catch some terraform syntax issues before waiting on ci.

PR #427, opened 2018-05-17
RDS cidr generation

jontours:
This one moderately more convoluted, but mostly same pattern here.


terraform/stacks/main/outputs.tf
```diff
@@ -193,10 +193,10 @@ output "rds_subnet_az2" {
     value = "${module.stack.rds_subnet_az2}"
 }
 output "rds_subnet_cidr_az1" {
-  value = "${var.rds_private_cidr_1}"
+  value = "${module.stack.rds_private_cidr_1}"
```
jmcarp: Looks like this isn't an output of the `stack` module--I think we'll need to add it.

ci/pipeline.yml
```diff
@@ -505,14 +503,12 @@ jobs:
       TF_VAR_az2: ((aws_az2))
       TF_VAR_rds_password: ((production_rds_password))
       TF_VAR_remote_state_bucket: ((aws_s3_tfstate_bucket))
-      TF_VAR_vpc_cidr: ((production_vpc_cidr))
+      TF_VAR_vpc_cidr: ((production_vpc_cidr))so
```
jmcarp: Typo?

ci/pipeline.yml
```diff
@@ -198,16 +198,18 @@ jobs:
       TF_VAR_aws_default_region: ((aws_default_region))
       TF_VAR_rds_password: ((tooling_rds_password))
       TF_VAR_concourse_prod_rds_password: ((concourse_prod_rds_password))
+      TF_VAR_concourse_prod_cidr: ((concourse_prod_cidr))
```
jmcarp: Did you mean to restore these params?

PR #457, opened 2018-06-29
Provision resources for CredHub across environments co-located with BOSH

jontours:
#Items so far added:
 - ~Load Balancer~
 - RDS


terraform/stacks/main/stack.tf
```diff
@@ -124,6 +124,26 @@ module "cf" {
     )}"]
 }
 
+module "credhub" {
+    source = "../../module/credhub"
+    
+    stack_description = "${var.stack_description}"
+
+    elb_main_cert_id = "${data.aws_iam_server_certificate.wildcard.arn}"
```
jmcarp: Looks like this variable isn't used in the `credhub` module. Can you drop any unused variables here?

terraform/stacks/main/stack.tf
```diff
@@ -124,6 +124,26 @@ module "cf" {
...
+    elb_apps_cert_id = "${data.aws_iam_server_certificate.wildcard_apps.arn}"
+    elb_subnets = ["${module.stack.public_subnet_az1}", "${module.stack.public_subnet_az2}"]
+    elb_security_groups = ["${var.force_restricted_network == "no" ?
+      module.stack.web_traffic_security_group :
+      module.stack.restricted_web_traffic_security_group}"]
+    
+    vpc_id = "${module.stack.vpc_id}"
+    vpc_cidr = "${var.vpc_cidr}"
+    rds_password = "${var.cf_rds_password}"
```
jmcarp: Looks like this will reuse the same rds password for the cf and credhub databases. Let's use a different password for each database.

ci/scripts/create-and-update-db.sh
```diff
@@ -33,11 +33,18 @@ if ! contains "$DATABASES" "$db_user"; then
...
 fi
 
+
+
 for db in ${DATABASES}; do
 
   # Create database
-  psql_adm -d postgres -l | awk '{print $1}' | grep -q "${db}" || \
-    psql_adm -d postgres -c "CREATE DATABASE ${db} OWNER ${db_user}"
+  if ! [ "${db}" = "uaadb" ]; then
```
rogeruiz: This would run for everything that isn't `uaadb` but that means it could run for other `postgres` databases besides CredHub right? I don't know if this is the best solution. I think we should just target `credhub` as a the database and do the specific stuff that we need to do there.

ci/scripts/create-and-update-db.sh
```diff
@@ -33,16 +33,24 @@ if ! contains "$DATABASES" "$db_user"; then
...
-    psql_adm -d "${db}" -c "CREATE EXTENSION IF NOT EXISTS \"${ext}\""
-  done
+  if ! [ "${db}" = "credhub" ]; then
+    psql_adm -d postgres -l | awk '{print $1}' | grep -q "${db}" || \
+      psql_adm -d postgres -c "CREATE DATABASE ${db} OWNER ${db_user}"
+    # Enable extensions
+    for ext in citext uuid-ossp pgcrypto pg_stat_statements; do
+      psql_adm -d "${db}" -c "CREATE EXTENSION IF NOT EXISTS \"${ext}\""
+    done
+  elif [ "${db}" = "credhub"]; then
```
rogeruiz: We don't need to do this sort of check. We should just have the `"${db}" = credhub` check and everything else gets the default which is the extension creation.

terraform/modules/credhub/variables.tf
```diff
@@ -0,0 +1,50 @@
+variable "stack_description" {}
+
+variable "credhub_cidr" {
```
jmcarp: This appears to be unused.

terraform/modules/credhub/variables.tf
```diff
@@ -0,0 +1,50 @@
...
+  default = "10.0.30.0/24"
+}
+
+variable "credhub_az" {
```
jmcarp: Also unused?

terraform/modules/credhub/variables.tf
```diff
@@ -0,0 +1,50 @@
...
+
+variable "rds_parameter_group_name" {
+  default = ""
+}
+
+variable "rds_db_size" {
+  default = 100
+}
+
+variable "rds_db_iops" {
```
jmcarp: Do we need to pay for provisioned iops here?

terraform/modules/credhub/variables.tf
```diff
@@ -0,0 +1,50 @@
...
+	default = 1000
+}
+
+variable "rds_db_storage_type" {
```
jmcarp: Is gp2 good enough?

terraform/modules/credhub/variables.tf
```diff
@@ -0,0 +1,50 @@
...
+  default = "io1"
+}
+
+variable "rds_instance_type" {
```
jmcarp: We don't have all that many secrets--think we can use a smaller instance type?

terraform/modules/credhub/variables.tf
rogeruiz: yeah, we were originally going to add a ELB to this, but we don't need it anymore good catch.

ci/scripts/update-credhub-db.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
jmcarp: Does this script need to exist? If flyway expects an empty database, do we have to do any of this?

jontours: The `create-and-update.sh` script adds some extensions.  When I did this work, I assumed those extensions were needed for compliance.  If that's not the case it would be _much_ easier to pull this script.

jmcarp: I don't know of a compliance need here. cc @LinuxBozo

LinuxBozo: No, there is no compliance need, there was however various other needs for either BOSH or CF so we just added them to both, but I don't know that Credhub has the same needs, though, it might when you know what each of these does.

citext: tolower string comparison in SQL queries
uuid-cissp: exactly how it sounds, for generating uuid
pgcrypto: for doing password hashing or other digests, like sha or md5
pg_stat_statements: keeping running stats of all queries that run

In reality, I have a feeling that pg_stat_statements is the one that is causing Credhub issues. So yeah, I would start with none, and then only add any that Credhub requires.

jontours: Yeah that was what our plan was for this.  From some research it seems that only `pg_stat_statements` and `citext` need some kind of state in the DB.  I was surprised by the `citext` needed some storage but from the docs it seems to be it needs storage for the operators.

rogeruiz:
Gotta update this to force SSL on the isntances

rogeruiz:
@jmcarp I'm good with the commits I've added on to this one. I think we can squash some too as I was experimenting with some settings here. That empty commit for instance can definitely be squashed.  Hows #468 going? can I change bases or can we merge those in first before merging this in? Also, this would create a database across all stacks. I only want to make a CredHub database in development for now. How should we proceed with this?

terraform/stacks/main/stack.tf
```diff
@@ -116,6 +116,17 @@ module "cf" {
     bucket_prefix = "${var.bucket_prefix}"
 }
 
+module "credhub" {
```
jmcarp: Didn't we say we wanted the credhub database to get created in the bosh module? cc @LinuxBozo

LinuxBozo: Since we've made the co-locating decision, yeah, that's where I would make it happen. The only reason to do the module separately is in case we want to eventually run a separate CH for Concourse at some point. Thoughts?

rogeruiz: I'm all for putting it where it makes sense today and then ripping it out if we decide to SaaS it later. Just unsure about where we put it if not in `modules/credhub` you mean `modules/bosh_vpc`?

terraform/modules/stack/base/base.tf
```diff
@@ -47,3 +47,20 @@ module "rds" {
   rds_subnet_group = "${module.rds_network.rds_subnet_group}"
   rds_security_groups = ["${module.rds_network.rds_postgres_security_group}"]
 }
+
+module "credhub_rds" {
+  source = "../rds"
```
jontours: I believe thi should be `../../rds`

rogeruiz:
RDS instances created previously outside of the `modules/stack/*` were deleted manually via the AWS console.

PR #466, opened 2018-07-19
[wip] Create private nat_gateways for AZ 1 & 2

rogeruiz:
Pairing with @sharms
Jira:
https://cm-jira.usa.gov/browse/CG-893

### DIFF
--- a/terraform/modules/bosh_vpc/network_private.tf
+++ b/terraform/modules/bosh_vpc/network_private.tf
@@ -135,3 +135,26 @@ resource "aws_instance" "az2_private_nat_2017_09" {
     Name = "${var.stack_description}  (AZ2 2017-09 NAT)"
   }
 }
+
+# AWS NAT Gateway provided by AWS
+resource "aws_nat_gateway" "az1_private_nat_gateway" {
+  allocation_id = "${aws_eip.nat.id}"
+  subnet_id     = "${aws_subnet.public.id}"
+
+  tags {
+    Name = "${var.stack_description}  (NAT Gateway AWS)"
+  }
+
+  depends_on = ["aws_internet_gateway.gw"]
+}
+
+resource "aws_nat_gateway" "az2_private_nat_gateway" {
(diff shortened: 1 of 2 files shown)


terraform/modules/bosh_vpc/network_private.tf
```diff
@@ -135,3 +135,26 @@ resource "aws_instance" "az2_private_nat_2017_09" {
     Name = "${var.stack_description}  (AZ2 2017-09 NAT)"
   }
 }
+
+# AWS NAT Gateway provided by AWS
+resource "aws_nat_gateway" "az1_private_nat_gateway" {
+  allocation_id = "${aws_eip.nat.id}"
```
jmcarp: I don't think either of these references exists--there's no `aws_eip` called `nat`, and there's no `aws_subnet` called `public`. And we'll want to use *different* EIPs and subnets for each NAT gateway.

rogeruiz: yeah these were just copied from the example for now. we figured we'd tackled this better in the morning on dev without rushing into things.

terraform/modules/bosh_vpc/network_private.tf
```diff
@@ -94,6 +94,27 @@ resource "aws_eip" "az2_nat_eip" {
...
+  instance = "${aws_instance.az2_private_nat_2017_09.id}"
+  vpc = true
+
+  count = "${var.use_nat_gateway_eip == "true" ? 1 : 0}"
+
+  lifecycle {
+    prevent_destroy = true
+  }
+}
+
```
rogeruiz: hmmm, not sure about this at all, since these EIPs are tied to instances and `aws_nat_gateway` is a whole 'nother thing.

jmcarp:
@rogeruiz let's hold off on this until we talk tomorrow. We're talking about a different order of operations in slack, so I think we're going to back out these changes.

PR #487, opened 2018-08-30
Update DMARC related TXT records

rogeruiz:
Fulfills request from GSAIT


rogeruiz:
Fulfills this request specifically:

> Since cloud.gov will be a non-mail sending domain, once you have made the
> changes to no longer use @cloud.gov emails, please work with Amazon to update
> the DNS as follows for cloud.gov, and let us know when it is complete...

PR #489, opened 2018-08-31
Add admin module to allow ip restriction

sharms:


terraform/modules/admin/admin_elb_main.tf
```diff
@@ -0,0 +1,32 @@
+resource "aws_lb_target_group" "admin" {
```
jmcarp: I'm guessing we're going to want to add the restricted security group here. But afaict security groups are associated with load balancers, not with target groups or listeners, so I think we'll have to add a new load balancer here too, then attach the security group.

PR #503, opened 2018-10-03
Bump aws to 1.39.0

jontours:
Have to test on dev then ready for review.


jontours:
Tested on development.

PR #529, opened 2018-12-29
Fix the path issues for $out_path

rogeruiz:
Seems like `$out_path` now includes a `README` file, so `cp` is having
issues finding the files. I've tested this in a Concourse worker to
determine that we should only be listing directories from the `live/`
directory and skip over any of the files in there such as `README`.

_This is brittle enough that it will break again if there's another
directory._

**This has been causing our certificate pipelines to fail**. This seems
unrelated to an issue that was brought up in #cg-support though around
certificates expiring.

### DIFF
--- a/ci/provision-certificate.sh
+++ b/ci/provision-certificate.sh
@@ -34,5 +34,5 @@ certbot certonly \
   --email "${EMAIL}" \
   --domain "${DOMAIN}"
 
-out_path=$(ls "${config_path}/live")
+out_path=$(ls -d -1 ${config_path}/live/*/)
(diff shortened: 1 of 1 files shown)


rogeruiz:
So, drama update on this, we've been provisioning certificates because of this, but failing to upload them. So our pipeline has been provisioning certs without updating them and then we hit the LetsEncrypt rate limit 😞

rogeruiz:
Well looks like this broke two months ago for us, but we weren't expiring so we never noticed. But Certbot added a `README` file to the output directory so that's why we ~~were~~ are having a bad time.

https://github.com/certbot/certbot/pull/6377

PR #554, opened 2019-04-18
Add access logs to all [AE]LBs

bengerman13:
To assist with troubleshooting, we need to turn these on for all the A- and E-LBs.
Note that with this change, all the logs go into the same bucket and directories, but the filenames are unique based on bucket, I.E.:
`cloud-gov-elb-logs/AWSLogs/<account number>/elasticloadbalancing/<region>/<year>/<month>/<day>/<account_number>_elasticloadbalancing_<region>_app.<ELB name>.<magic string>_<timestamp>_<ip address>_<more magic strings>.log.gz`

### DIFF
--- a/terraform/modules/diego/diego_elb_main.tf
+++ b/terraform/modules/diego/diego_elb_main.tf
@@ -81,4 +81,8 @@ resource "aws_elb" "diego_elb_main" {
   tags {
     Name = "${var.stack_description}-Diego-Proxy-ELB"
   }
+
+   access_logs = {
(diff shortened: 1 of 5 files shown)


bengerman13:
> So...I want this, and I want this really badly. I also want this to be manageable. I'm going to approve it because it solves a non-trivial amount of problems, but I want to have a discussion on how we can ensure these logs end up in logsearch/log aggregation so they're searchable.

Agreed - the first step to get logs into logsearch is going to be getting them into s3 no matter what 😄

PR #570, opened 2019-06-05
Logging organization

bengerman13:
picture of Ben at work

This PR creates a logging bucket, allows concourse workers to write to it, and configures all of the A/ELBs to write logs to it. 

I'm very unsure about whether I've declared/defined/passed variables around in too many/few places.

### DIFF
--- a/terraform/modules/log_bucket/log_bucket.tf
+++ b/terraform/modules/log_bucket/log_bucket.tf
@@ -1,0 +1,66 @@
+locals {
+# https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#access-logging-bucket-permissions
+    aws_alb_account_ids = {
+        us-east-1 = "127311923021",
+        us-east-2 = "033677994240",
+        us-west-1 = "027434742980",
+        us-west-2 = "797873946194",
+        ca-central-1 = "985666609251",
+        eu-central-1 = "054676820928",
+        eu-west-1 = "156460612806",
+        eu-west-2 = "652711504416",
+        eu-west-3 = "009996457667",
+        eu-north-1 = "897822967062",
+        ap-east-1 = "754344448648",
+        ap-northeast-1 = "582318560864",
+        ap-northeast-2 = "600734575887",
+        ap-northeast-3 = "383597477331",
+        ap-southeast-1 = "114774131450",
+        ap-southeast-2 = "783225319266",
+        ap-south-1 = "718504428378",
+        sa-east-1 = "507241528517",
+        us-gov-west-1 = "048591011584",
+        us-gov-east-1 = "190560391635",
+        cn-north-1 = "638102146993",
+        cn-northwest-1 = "037604701340"
+    }
+}
+resource "aws_s3_bucket" "log_bucket" {
+    bucket = "${var.log_bucket_name}"
+    acl = "${var.log_bucket_acl}"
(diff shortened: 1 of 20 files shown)


terraform/modules/log_bucket/log_bucket.tf
```diff
@@ -0,0 +1,66 @@
+locals {
+# https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-access-logs.html#access-logging-bucket-permissions
+    aws_alb_account_ids = {
```
mxplusb: I don't think we need non-govcloud account ids, yeah?

terraform/modules/log_bucket/log_bucket.tf
```diff
@@ -0,0 +1,66 @@
...
+    force_destroy = "${var.log_bucket_force_destroy}"
+
+    lifecycle_rule {
+        prefix = ""
+        enabled = true
+        transition {
+            days = 90
+            storage_class = "ONEZONE_IA"
+        }
+        transition {
```
mxplusb: Do we want to keep logs past 180 days? Glacier is great for storing logs but if we ever had to access them, it's egregiously expensive.

terraform/modules/log_bucket/log_bucket.tf
bengerman13: meh. I copy-pasted and formatted the whole table. No harm in it being here, as long as it's accurate.

terraform/modules/log_bucket/log_bucket.tf
bengerman13: this was a shot in the dark - currently, we keep all of them, forever, which seems excessive.
I'm not opposed to dropping the glacier step.
Do we want to just do one zone IA at 90 then delete at 180? 
@jontours any thoughts?

jontours: I'm fine with it as is, I'd rather not optimize for something that's not even been used yet.  But I know zero glacier.

mxplusb: > Do we want to just do one zone IA at 90 then delete at 180?

I like this.

terraform/modules/log_bucket/log_bucket.tf
mxplusb: Let's only keep the govcloud ones, just to be safe.

bengerman13: I don't see added safety in removing them, since they're keyed off of the aws region we're in.

mxplusb: Under the doc page you linked, this was the first requirement:

> The bucket must be located in the same region as the load balancer.

We cannot deploy to any of the other regions and since we can't deploy to those regions, we can't use it, which means we're putting extraneous account principals into our bucket access policy.

bengerman13: that's *exactly* why I think it's valuable to keep these here. 
Our contingency plan includes the possibility of rolling to a non-govcloud AWS region.

Again, we're indexing into this map based on the region we're deploying to. The only way we'll access the non-govcloud regions here is if the rest of the deployment is going to a non-govcloud region. This module should not be the one that decides which regions are/are not allowed.

mxplusb: You're right, I missed it:

```
aws_alb_account_ids[var.aws_partition]
```

PR #622, opened 2019-11-13
added domain broker user in AWS.

mxplusb:
adds a dedicated aws user and access key for the domain broker, and sets the policy for it.

### DIFF
--- a/terraform/stacks/main/domains_broker.tf
+++ b/terraform/stacks/main/domains_broker.tf
@@ -203,6 +203,57 @@ output "domains_broker_listener_arns" {
 }
 
 /* new broker ALB */
+
+resource "aws_iam_user" "domain_broker_v2" {
+  name = "domain_broker_v2"
+}
+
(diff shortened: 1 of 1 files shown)


terraform/stacks/main/domains_broker.tf
```diff
@@ -203,6 +203,34 @@ output "domains_broker_listener_arns" {
...
+  policy = <<EOF
+  {
+    "Version": "2015-12-01",
+    "Statement": [
+      {
+        "Action": [
+          "elasticloadbalancing:*"
+        ],
+        "Effect": "Allow",
+        "Resource": "*"
```
bengerman13: this seems overly-broad. Can we narrow it down, at least to the names of the alb's it's expected to manage?

mxplusb: Yes, I am very on board with that. And of course I think I just found it. Bleh, I couldn't figure it out originally, lol.

PR #636, opened 2019-11-25
change domain broker from cloud.gov hz to cdn-broker-test.cloud.gov

mxplusb:
This was stuck in a resolving loop, by using the CDN broker hosted zone, I can create test CNAME records, resolving to the ELBs, which send the traffic from the other domain through the platform. This is the proper design, the existing LE design I was using was wrong.

### DIFF
--- a/terraform/stacks/external/stack.tf
+++ b/terraform/stacks/external/stack.tf
@@ -33,8 +33,8 @@ module "lets_encrypt_user" {
 }
 
 module "domain_broker_v2_user" {
-  source = "../../modules/iam_user/lets_encrypt"
+  source = "../../modules/cdn_broker"
   aws_partition = "${data.aws_partition.current.partition}"
-  hosted_zone = "${var.lets_encrypt_hosted_zone}"
-  username = "domain-broker-v2-${var.stack_description}"
+  hosted_zone = "${var.cdn_hosted_zone}"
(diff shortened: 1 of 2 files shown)


mxplusb:
I think I have to add the `cdn_hosted_zone` variable reference somewhere (i.e. so terraform knows which hosted zone to use), but I'm not quite sure where just yet.

terraform/stacks/external/stack.tf
```diff
@@ -33,8 +33,8 @@ module "lets_encrypt_user" {
...
 module "domain_broker_v2_user" {
-  source = "../../modules/iam_user/lets_encrypt"
+  source = "../../modules/cdn_broker"
   aws_partition = "${data.aws_partition.current.partition}"
-  hosted_zone = "${var.lets_encrypt_hosted_zone}"
-  username = "domain-broker-v2-${var.stack_description}"
-}
\ No newline at end of file
+  hosted_zone = "${var.cdn_hosted_zone}"
+  username = "domain-broker-v2"
```
bengerman13: this seems like we'll only have one user, shared between dev, staging and prod. Is that correct?

mxplusb: Yes, I'm just following the pre-existing pattern, but one for each env would be good. What do you want to do?

PR #654, opened 2020-01-17
updated concourse worker policy.

mxplusb:
Allowing policy that allows for Concourse to read it's own pipeline credentials.

Reference build: [#1800](https://ci.fr.cloud.gov/teams/main/pipelines/terraform-provision/jobs/plan-bootstrap-tooling/builds/1800)

## Security Considerations

Concourse can read but not write credentials from a variable store without an operator having to manually set them in the pipeline. This doesn't change the existing posture, just streamlines the process.


bengerman13:
This looks fine to me, but there are a bunch of changes that have built up in TF (I think from the work @ChrisMcGowan and @soutenniza have been working on), so shipping it might get weird

PR #655, opened 2020-01-17
tf: version upgrade

mxplusb:
Upgrade to the latest non-beta version of the aws terraform provider and update the mappings as well.

## Security Considerations

None that I can think of.

## Builds

DNS: [#997](https://ci.fr.cloud.gov/teams/main/pipelines/terraform-provision/jobs/plan-dns/builds/997)
Staging: [#1735](https://ci.fr.cloud.gov/teams/main/pipelines/terraform-provision/jobs/plan-bootstrap-staging/builds/1735)
Tooling: [#1825](https://ci.fr.cloud.gov/teams/main/pipelines/terraform-provision/jobs/plan-bootstrap-tooling/builds/1825)
Production: [#1725](https://ci.fr.cloud.gov/teams/main/pipelines/terraform-provision/jobs/plan-bootstrap-production/builds/1725)
External Staging: [#1676](https://ci.fr.cloud.gov/teams/main/pipelines/terraform-provision/jobs/plan-bootstrap-external-staging/builds/1676)
External Production: [#1653](https://ci.fr.cloud.gov/teams/main/pipelines/terraform-provision/jobs/plan-bootstrap-external-production/builds/1653)

Signed-Off-By: Mike Lloyd <mike.lloyd@gsa.gov>


terraform/modules/admin/admin_elb_main.tf
```diff
@@ -51,7 +51,8 @@ resource "aws_lb_listener_rule" "admin" {
   }
 
   condition {
-    field  = "host-header"
-    values = ["${element(var.hosts, count.index)}"]
+    host_header {
+      values = ["${element(var.hosts, count.index)}"]
```
tammersaleh: we can simplify this as:

``` hcl
values = [element(var.hosts, count.index)]
```

tammersaleh: In fact, a `%s/"$(\(.*\))"/\1/` in vim would catch a lot of these, but will also accidentally convert `"$(foo).$(bar)"` to `foo).(bar`

tammersaleh: also, for this example, we could use the new [`for_each`](https://www.terraform.io/docs/configuration/resources.html#for_each-multiple-resource-instances-defined-by-a-map-or-set-of-strings) instead of `count` - that would eliminate the need for `element()`

mxplusb: @tammersaleh you're thinking of the 0.12 upgrade, we're still using 0.11, the 0.12 upgrade is not a part of this change, this is just an upgrade to the provider.

PR #661, opened 2020-01-28
updated the limit checker policy with the latest version.

mxplusb:
[Reference template](https://awslimitchecker.readthedocs.io/en/latest/iam_policy.html). It looks like the original template was copied from there, so this just syncs it with reality.

## Security Considerations

It's an extension of permissions but readonly, so still safe.


bengerman13:
Concourse thinks you've got more to do here

PR #675, opened 2020-02-07
Update documentation

tammersaleh:

* The README is now concise and up to date
* The bootsrapping and validating bits have been moved to separate files

## Security considerations

None.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,126 +1,86 @@
-# cg-provision
+# Cloud.gov Provisioning System
 
-Scripts, configurations, and procedures for provisioning the infrastructure to set up cloud.gov.
+This repository holds the terraform configuration (and BOSH vars and ops-files)
+to bootstrap our infrastructure.
 
-**Install these first on your laptop:**
-* git
-* [Terraform](https://www.terraform.io/)
-* the [AWS Command Line Interface (CLI) tool](https://aws.amazon.com/cli/)
-*  [`jq`, a command line JSON processor](https://stedolan.github.io/jq/)
-* the [bosh cli](https://bosh.io/docs/cli-v2.html)
+## Layout
 
-macOS users can install all of these with [`homebrew`](https://brew.sh/).
+### Terraform
 
-## Bootstrap procedure
+This system is using Terraform 0.11.  Future versions of the main providers
+(AWS in particular) [will not work with this
+version](https://www.hashicorp.com/blog/deprecating-terraform-0-11-support-in-terraform-providers/).
+At some point, we'll need to upgrade to the (much better) 0.12 syntax.
 
-Follow this procedure to setup a bootstrap instance of concourse and deploy minimal components to bring up a permanant instance of concourse deployed by a bosh director.
+The main terraform directories are:
 
-1. Clone this repository: `git clone https://github.com/18F/cg-provision`
-1. Get the rest of the necessary repositories: `./cg-provision/scripts/bootstrap/setup-bootstrap.sh`
-1. `cd cg-provision`
-1. Make a copy of `env.example.sh` and populate with AWS credentials, etc. (`cp env.example.sh env.sh`)
-    1. An example `TERRAFORM_PROVISION_CREDENTIALS_FILE` can be found in `ci/credentials.example.yml`. Make a copy and place in your `${WORKSPACE_DIR}` from `env.sh`.
-    1. `toolingbosh` and `concourse` currently need [pre-populated, encrypted secrets files](https://docs.cloud.gov/ops/updating-cf/#updating-secrets-yml): `tooling-bosh-main.yml`,`tooling-bosh-external.yml`,`concourse-tooling-prod.yml`.
-        1. Populate these and encrypt with `../cg-pipeline-tasks/encrypt.sh`, upload to `${VARZ_BUCKET}` with the aws cli, and set the respective passphrases in `env.sh`.
-        1. **TODO: generate all secrets for bosh & concourse / pull values from tf.**
-    1. `source env.sh`
-    1. If the above step fails, you may have to comment out the `TF_STATE_BUCKET` line.
-1. Create bootstrap terraform stack: `./scripts/bootstrap/01-bootstrap-terraform.sh`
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,126 +1,86 @@
...
+to bootstrap our infrastructure.
+
+## Layout
+
+### Terraform
+
+This system is using Terraform 0.11.  Future versions of the main providers
+(AWS in particular) [will not work with this
+version](https://www.hashicorp.com/blog/deprecating-terraform-0-11-support-in-terraform-providers/).
+At some point, we'll need to upgrade to the (much better) 0.12 syntax.
```
mxplusb: This is not a good place to put this information - it's best to capture it as a story elsewhere.

README.md
```diff
@@ -1,126 +1,86 @@
...
+* `stacks`: the various "environments"
+
+#### Environments
+
+The `main` stack is a template that is used to provision the production,
+staging, and development "environments."
+
+> ##### Aside: Multiple environments in one account
+>
+> Why do I keep putting "environments" in quotes?  One confusing and suboptimal
```
mxplusb: I think this can live somewhere else, we don't want this information being broadcast while we're in this state.

README.md
```diff
@@ -1,126 +1,86 @@
...
+> AWS account.
+>
+> :shrug:
+
+The `tooling` stack contains our "proto-BOSH", which deploys the tooling BOSH.
+The tooling BOSH then deploys the BOSH directors in the main stacks.
+
+The `external` and `dns` stacks are both outside of GovCloud (commercial AWS).
+
+I believe the `test` stack is unused.
```
mxplusb: It is, it would be good to put in a story to delete it and then remove this doc bit.

README.md
```diff
@@ -1,126 +1,86 @@
...
+
+### BOSH
+
+The `bosh` directory contains vars and opsfiles.  I'm not sure which directors
```
mxplusb:
```suggestion
The `bosh` directory contains vars and opsfiles.
```

README.md
```diff
@@ -1,126 +1,86 @@
...
+1. Create a branch and pull-request with your changes and ask for review and
+   merge from a teammate.
+1. Once the teammate :thumbsup: the changes, head over to the Concourse
+   pipeline and review the resultant Terraform plan output.
+1. If the plan looks like what you intended, then manually trigger the
+   appropriate apply jobs.
+
+## Other Points of Note
+
+You may see `access_key_id_prev` and `aws_key_id_prev` as outputs for our `iam`
```
mxplusb: This should likely go in our other docs.

README.md
tammersaleh: I see your point, but I'm not sure obscuring this fact will outweigh the benefit of making the configuration clear to the engineer tasked with working in this repository.  How strongly do you feel about this?

README.md
tammersaleh: Rather than remove this line, do you know which directors use the ops and vars files?  If so, I'll just put that in the docs.

README.md
tammersaleh: It's actually already in our other docs, and linked to in the README here.  I feel it's important to have pointers like these in the developer documentation to aid discoverability.  In this case, I was very confused by terraform outputs that were hardcoded to empty strings.  It took me a while to find the reason in our public documentation, so having this pointer here helps the next person avoid that detective work.

I'd also be happy to move this to a comment above the various `*_prev` outputs.

README.md
tammersaleh: I put a story in place and linked it here.

README.md
tammersaleh: You're right - removed

README.md
mxplusb: Not without looking at the pipeline config. I would say unless you feel like going through that effort to document it, there's no point in commenting on individual uncertainty in a readme.

README.md
mxplusb: Very strongly. This type of commentary belongs in internal documentation where we can debate it freely, not public readmes.

README.md
mxplusb: Not without looking at the pipeline config. I would say unless you feel like going through that effort to document it, there's no point in commenting on individual uncertainty in a readme.

README.md
mxplusb: Sorry, I wasn't clear (my fault), this should go in our internal documentation, there's no real need to highlight output and where to find it for important things like AWS keys in public documentation.

bootstrap_and_teardown.md
```diff
@@ -0,0 +1,93 @@
+# Bootstrap procedure
```
mxplusb: This file was just renamed, yeah?

validating_locally.md
```diff
@@ -0,0 +1,22 @@
+## Validating `terraform plan` locally (beta)
```
mxplusb: Maybe work in progress instead of beta? I'm on the fence about running terraform tasks locally beyond `terraform validate` due to the visibility of state. Or is that what this file is doing and it just needs to be captured more clearly?

README.md
bengerman13: I think this belongs here - it's specific to how this repository is structured and does not at all impact security

bootstrap_and_teardown.md
tammersaleh: This file was extracted from the original `README.md`

validating_locally.md
tammersaleh: This file was extracted from the original `README.md`, but I'm happy to change that text.

README.md
tammersaleh: Done

README.md
```diff
@@ -1,126 +1,63 @@
...
+
+The `main` stack is a template that is used to provision the production,
+staging, and development "environments."
+
+The `tooling` stack contains our "proto-BOSH", which deploys the tooling BOSH.
+The tooling BOSH then deploys the BOSH directors in the main stacks.
+
+The `external` and `dns` stacks are both outside of GovCloud (commercial AWS).
+
+I believe the `test` stack is unused, and there's [a story to remove
```
mxplusb:
```suggestion
The `test` stack is unused, and there's [a story to remove
```

PR #697, opened 2020-03-04
add policy for self-managing MFA

bengerman13:
- add a self-managed MFA policy

## security considerations
None, directly. Adding this policy to users will allow them to enable MFA on their accounts, which is good.

### DIFF
--- a/terraform/modules/iam_role_policy/self_managed_mfa/policy.json
+++ b/terraform/modules/iam_role_policy/self_managed_mfa/policy.json
@@ -1,0 +1,49 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+      {
+          "Sid": "AllowViewAccountInfo",
+          "Effect": "Allow",
+          "Action": "iam:ListVirtualMFADevices",
+          "Resource": "*"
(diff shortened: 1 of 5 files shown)


terraform/modules/iam_role_policy/self_managed_mfa/policy.json
```diff
@@ -0,0 +1,49 @@
...
+              "iam:DeactivateMFADevice",
+              "iam:EnableMFADevice",
+              "iam:GetUser",
+              "iam:ListMFADevices",
+              "iam:ResyncMFADevice"
+          ],
+          "Resource": "arn:${aws_partition}:iam::*:user/$${aws:username}"
+      },
+      {
+          "Sid": "DenyAllExceptListedIfNoMFA",
```
tammersaleh: So this clause means people can't do _anything_ until they configure MFA.  Cool if that's what we want, but should be aware.

mxplusb: This is fine by me, but good call out that we should document it.

PR #703, opened 2020-03-13
Create `cg-build-artifacts` bucket for use in concourse pipelines

tammersaleh:

Add a `cg-build-artifacts` bucket for use in concourse pipelines.  This bucket allows us to throw cached assets in s3 that are created early in the pipeline, but required later during deploy.  A good example is the output from `node build`.

## security considerations

Concourse workers have read/write access to this bucket.  This is in line with our other buckets.

### DIFF
--- a/terraform/stacks/tooling/buckets.tf
+++ b/terraform/stacks/tooling/buckets.tf
@@ -52,7 +52,7 @@ module "log_bucket" {
   aws_region      = "${data.aws_region.current.name}"
 }
 
-module "billing_bucket_staging" {
+module "build_artifacts_bucket" {
   source        = "../../modules/s3_bucket/encrypted_bucket"
   bucket        = "${var.bucket_prefix}build-artifacts"
   aws_partition = "${data.aws_partition.current.partition}"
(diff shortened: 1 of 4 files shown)


tammersaleh:
My editor automatically runs `terraform fmt` on write.  Let me know if this is just annoying, and I can undo those whitespace changes.

terraform/stacks/tooling/buckets.tf
```diff
@@ -1,53 +1,59 @@
...
-  source = "../../modules/log_bucket"
-  aws_partition = "${data.aws_partition.current.partition}"
+  source          = "../../modules/log_bucket"
+  aws_partition   = "${data.aws_partition.current.partition}"
   log_bucket_name = "${var.log_bucket_name}"
-  aws_region = "${data.aws_region.current.name}"
+  aws_region      = "${data.aws_region.current.name}"
+}
+
+module "billing_bucket_staging" {
```
bengerman13: this should probably have a unique, descriptive name, no?

tammersaleh: If you wanna be picky about it.

PR #708, opened 2020-03-14
Enable versioning on the cg-build-artifacts bucket

tammersaleh:

Enable versioning on the cg-build-artifacts bucket

## security considerations

None


tammersaleh:
@mxplusb This bucket is primarily for the concourse s3 resource, which says:

>  If you enable versioning for your S3 bucket then you can keep the file name the same and upload new versions of your file without resorting to version numbers.

Also, just seems like best practice to enable versioning on all S3 buckets 🤷‍♀

tammersaleh:
As per discussion with @mxplusb and @bengerman13 - Holding off on merging this until after a site launch.

PR #713, opened 2020-04-02
Don't create foreign key, since it fails

bengerman13:
For some reason, username is not unique in uaadb, causing the fkey relationship from totp_seed to fail. Skip creating this constraint so the table will create.

I poked around the existing databases, and this matches reality.

- Skip creating a foreign key on totp_seed.username

## security considerations
None


spgreenberg:
username is not unique b/c the same user can have multiple origins.

PR #720, opened 2020-04-20
Additional perms needed to view WAF

pburkholder:

- create a new iam_role_policy `compliance-role`
- this is to be used in addition to the aws-managed `Security Audit` role
- It is OK to use this policy in both Commercial and GovCloud. Actions that aren't applicable will be ignored (the `Security Audit` roles does the same thing).
- Denies the few `GetToken..` actions that seem dicey.

## security considerations

Gives more read-only permission to compliance role

### DIFF
--- a/terraform/modules/iam_role_policy/compliance_role/outputs.tf
+++ b/terraform/modules/iam_role_policy/compliance_role/outputs.tf
@@ -1,0 +1,6 @@
+output "name" {
+  value = "${aws_iam_policy.iam_policy.name}"
+}
+output "arn" {
+  value = "${aws_iam_policy.iam_policy.arn}"
+}
--- a/terraform/modules/iam_role_policy/compliance_role/policy.json
+++ b/terraform/modules/iam_role_policy/compliance_role/policy.json
@@ -9,23 +9,20 @@
             "waf:List*",
             "waf:Get*",
             "waf-regional:List*",
-            "waf-regional:Get*"
+            "waf-regional:GetByteMatchSet",
+            "waf-regional:GetIPSet",
+            "waf-regional:GetRule",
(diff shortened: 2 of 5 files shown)


spgreenberg:
What is the purpose of the "deny" role?

pburkholder:
@spgreenberg There are 11 `Get*` actions for WAF (e.g. https://iam.cloudonaut.io/reference/waf-regional.html) and 9 of them seem useful and unrelated to anything that could change our WAF. The other two are 

```
waf-regional:GetChangeToken | 
When you want to create, update, or delete AWS WAF objects, get a change token and include the change token in the create, update, or delete request.
``` 
```
waf-regional:GetChangeTokenStatus | 
Returns the status of a ChangeToken that you got by calling GetChangeToken.
```



So I figured my options were:

1. Allow `Get*` and not worry about the GetChangeToken, or GetChangeTokenStatus in read-only/auditor role
2. Allow `Get*` and deny the GetChangeToken and GetChangeTokenStatus actions
3.  Allow each of 9 individual Gets that are of interest

So I chose option 2., but I may be doing it wrong, and 1 or 3 is a better path.

bengerman13:
I think 3 is the best path here, as it's more clear (IMO) and won't automatically gain new permissions as AWS adds them

ccostino:
I agree with @bengerman13 is right, if another role is added that starts with `Get*` it'll automatically be included, even if it's something of a `GetChangeToken` nature.  Better to be explicit in this case to prevent any future surprises!

pburkholder:
Cool, I've made another commit to address that.

The `wafv2` API seems to be much better specified in that the Get*, List*, and Describe* actions are all read-only, so I hope that's still OK w/ y'all.

PR #721, opened 2020-04-21
add policy for self-managed access keys

bengerman13:
- add self-managed access keys policy to AWS

## security considerations
None directly. This merely creates the policy. Applying them later, of course, allows access to these resources


pburkholder:
The name might be more accurately `self-managed-credentials` but LGTM.

Notes:
It looks like this policy comes from https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_aws_my-sec-creds-self-manage-pass-accesskeys-ssh.html . 
The AWS docs also reference this policy:https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_iam_credentials_console.html which includes "iam:*ServiceSpecificCredential*", and  "iam:*SigningCertificate*"

bengerman13:
Would it be better, maybe, to collapse this all into one policy that allows users to:
- update console passwords
- update ssh keys
- update api keys
- update signing certificates
- update service-specific credentials (currently these are just for CodeCommit)
- update MFA

I don't currently see a need for a policy that lets you update one type of credential but not another

PR #724, opened 2020-05-14
add payer account access to read S3 bucket tagging

bengerman13:
- allow payer account to read s3 bucket tags

## security considerations
Allows users of the payer account to see the names and tags of all of our buckets

### DIFF
--- a/terraform/stacks/tooling/cross_account_iam.tf
+++ b/terraform/stacks/tooling/cross_account_iam.tf
@@ -41,7 +41,7 @@ resource "aws_iam_policy" "read_s3_tags" {
             "s3:GetBucketLocation",
             "s3:GetBucketTagging"
            ],
-          "Resource": "arn:aws:s3:::*"
+          "Resource": "arn:${data.aws_partition.current.partition}:s3:::*"
         }
       ]
     }
--- a/terraform/stacks/tooling/cross_account_iam.tf
+++ b/terraform/stacks/tooling/cross_account_iam.tf
@@ -1,0 +1,54 @@
+resource "aws_iam_role" "payer_account_access" {
+    name = "payer-account-access"
+
+    assume_role_policy = <<EOF
+    {
+      "Version": "2012-10-17",
+      "Statement": [
+        {
+          "Effect": "Allow",
+          "Principal": {
+            "AWS": "arn:aws:iam::${var.payer_account_id}:root"
+          },
+          "Action": "sts:AssumeRole",
+          "Condition": {
+            "Bool": {
+              "aws:MultiFactorAuthPresent": "true"
+            }
+          }
+        }
+      ]
+    }
+    EOF
+}
+
+resource "aws_iam_policy" "read_s3_tags" {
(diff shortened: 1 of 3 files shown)


terraform/stacks/tooling/cross_account_iam.tf
```diff
@@ -0,0 +1,54 @@
...
+          "Action": [
+            "s3:GetBucketLocation",
+            "s3:GetBucketTagging"
+           ],
+          "Resource": "arn:aws:s3:::*"
+        }
+      ]
+    }
+    EOF
+}
```
ccostino, line 49: Just confirming:  the payer account only needs to read S3 bucket tags and that's it?  No need for writing tags or anything else?

bengerman13: yeah - the payer account needs to enable cost allocation tracking on some tags we're applying. The trick is the cost allocation tracking is a feature of their account, but relies on seeing the tags in our account.

terraform/stacks/tooling/cross_account_iam.tf
```diff
@@ -0,0 +1,54 @@
...
+          "Action": "s3:ListAllMyBuckets",
+          "Resource": "*"
+        },
+        {
+          "Effect": "Allow",
+          "Action": [
+            "s3:GetBucketLocation",
+            "s3:GetBucketTagging"
+           ],
+          "Resource": "arn:aws:s3:::*"
```
soutenniza, line 44: Is this for aws commerical only? If not may want to use something so arn is correct for aws gov cloud.

Similar to this https://github.com/cloud-gov/cg-provision/blob/payer-read-access/terraform/stacks/tooling/s3-logs.tf#L25

bengerman13: Good catch! thanks @soutenniza !

terraform/stacks/tooling/cross_account_iam.tf
```diff
@@ -0,0 +1,54 @@
...
+    name = "payer-account-access"
+
+    assume_role_policy = <<EOF
+    {
+      "Version": "2012-10-17",
+      "Statement": [
+        {
+          "Effect": "Allow",
+          "Principal": {
+            "AWS": "arn:aws:iam::${var.payer_account_id}:root"
```
soutenniza, line 11: Probably same issue here :)

bengerman13: I don't think so here - here we know they are a commercial account, since they're the payer :)

soutenniza: Ok, sounds good then

pburkholder: Looks like you're trying to delegate access between an account in the `aws` partition and the `aws-us-gov` partition, which you cannot do :(.   https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html : "You cannot delegate access between accounts in different partitions."

PR #736, opened 2020-05-18
Add support for "sts:GetFederationToken"

ccostino:
This change adds support for `sts:GetFederationToken` to the self-managed-credentials policy so that users are able to use `aws-vault login` to get to the AWS web console.

This action had to be added to two different sections to account for the different levels of access our team has.

- Add `sts:GetFederationToken` to the `DenyAllExceptListedIfNoMFA` section

## Security considerations
We need to consider the impact of [`sts:GetFederationToken`](https://docs.aws.amazon.com/STS/latest/APIReference/API_GetFederationToken.html) for anyone who has the `self-managed-credentials` policy attached.

### DIFF
--- a/terraform/modules/iam_role_policy/self_managed_credentials/policy.json
+++ b/terraform/modules/iam_role_policy/self_managed_credentials/policy.json
@@ -8,8 +8,7 @@
                 "iam:GetAccountPasswordPolicy",
                 "iam:ListVirtualMFADevices",
                 "iam:ListUsers",
-                "iam:GetAccountSummary",
-                "sts:GetFederationToken"
+                "iam:GetAccountSummary"
             ],
             "Resource": "*"
         },
--- a/terraform/modules/iam_role_policy/self_managed_credentials/policy.json
+++ b/terraform/modules/iam_role_policy/self_managed_credentials/policy.json
@@ -8,7 +8,8 @@
                 "iam:GetAccountPasswordPolicy",
                 "iam:ListVirtualMFADevices",
                 "iam:ListUsers",
-                "iam:GetAccountSummary"
+                "iam:GetAccountSummary",
+                "sts:GetFederationToken"
             ],
             "Resource": "*"
         },
@@ -70,7 +71,8 @@
               "iam:ListMFADevices",
               "iam:ListVirtualMFADevices",
               "iam:ResyncMFADevice",
-              "sts:GetSessionToken"
+              "sts:GetSessionToken",
+              "sts:GetFederationToken"
           ],
           "Resource": "*",
           "Condition": {


ChrisMcGowan:
So in reading this: https://docs.aws.amazon.com/STS/latest/APIReference/API_GetFederationToken.html

Two things I see in this statement on that page:

> Permissions
> 
> You can use the temporary credentials created by GetFederationToken in any AWS service except the following:
> 
> You cannot call any IAM operations using the AWS CLI or the AWS API.
> 
> You cannot call any STS operations except GetCallerIdentity

@ccostino @pburkholder - i'm not sure if anyone would be doing IAM ops via CLI and based on the 2nd statement not sure how "sts:GetSessionToken" would work

ccostino:
Thanks, @ChrisMcGowan!  I don't foresee anyone working with IAM in the CLI either, and are you referring to the fact that [`sts:GetSessionToken`](https://docs.aws.amazon.com/STS/latest/APIReference/API_GetSessionToken.html) is already listed in the policy and therefore may not be usable given the documentation?

I think we're okay here based on the documentation; it seems that both of these STS operations are used to enable access depending on what the user is trying to do.  It seems `sts:GetSessionToken` is meant more for API calls and presumably the CLI, while `sts:GetFederationToken` enables the ability to login to the console via the CLI (in fact, the reason why we added it was due to the specific error that we were seeing, the operation was explicitly being denied due to `sts:GetFederationToken` missing).

I see that `sts:GetSessionToken` has similar caveats as `sts:GetFederationToken` regarding permissions:

>Permissions
>
>The temporary security credentials created by GetSessionToken can be used to make API calls to any AWS service with the following exceptions:
>
>You cannot call any IAM API operations unless MFA authentication information is included in the request.
>
>You cannot call any STS API except AssumeRole or GetCallerIdentity.

@pburkholder is my assessment correct in the case of `sts:GetSessionToken` and why it was already in the policy?  If so then there shouldn't be an issue with the two of these conflicting or interacting with each other.

We had also tested the existing functionality of running commands and therefore assuming a role/getting temporary credentials and then using `aws-vault login`, both had worked with this change.

terraform/modules/iam_role_policy/self_managed_credentials/policy.json
```diff
@@ -8,7 +8,8 @@
                 "iam:GetAccountPasswordPolicy",
                 "iam:ListVirtualMFADevices",
                 "iam:ListUsers",
-                "iam:GetAccountSummary"
+                "iam:GetAccountSummary",
+                "sts:GetFederationToken"
```
pburkholder, line 12: Let's remove this. We're not quite sure of the impact when the `Resource` is `*`.  It only affects non-admins.  I can live without it (Or when I can't live without it,  I can research the correct Resource to use).

pburkholder:
I think this is what needs to be allowed if we want to be strict:

```
{
  "Version": "2012-10-17",
  "Statement": [{
    "Effect": "Allow",
    "Action": "sts:GetFederationToken",
   "Resource": "arn:${aws_partition}:iam::*:federated-user/$${aws:username}
}
```

PR #742, opened 2020-05-29
Simplify configuration for external-domain-broker

tammersaleh:

- We should NOT be passing in `TF_VAR_foo_staging` or `TF_VAR_foo_production` variables into our pipeline tasks.  We already pass in `STACK_NAME`.  This should be enough.
- Change values to match what's configured in the broker and its pipeline.

[Closes cloud-gov/external-domain-broker#71]

## security considerations

None

### DIFF
--- a/terraform/modules/external_domain_broker/resources.tf
+++ b/terraform/modules/external_domain_broker/resources.tf
@@ -1,5 +1,5 @@
 resource "aws_route53_zone" "zone" {
-  name    = "${var.hosted_zone}"
+  name    = "external-domains-${var.stack_description}.cloud.gov"
   comment = "Hosts TXT and CNAME records for the external-domain-broker"
 }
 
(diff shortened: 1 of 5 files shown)


tammersaleh:
@bengerman13 I'd like to walk through this with you when you have time.

terraform/modules/external_domain_broker/resources.tf
```diff
@@ -1,5 +1,5 @@
 resource "aws_route53_zone" "zone" {
-  name    = "${var.hosted_zone}"
+  name    = "external-domains-${var.stack_description}.cloud.gov"
```
ccostino, line 2: Noting that while pairing during a walkthrough review of this, we noticed the name was wrong and needed to be fixed to start with `external-domains` (the `s` was missing).

It's already fixed here as a part of the review.

PR #746, opened 2020-06-03
New users need ChangePassword before enabling MFA

pburkholder:
Working with @onelittlebecca we realized that her temporary password has to be changed on her first login, before she's set up an MFA device. So we need newly onboarded users to ChangePassword.

Screenshot

image

- one line policy change

## security considerations

We need to discuss if this is change would allow MFA to be bypassed

### DIFF
--- a/terraform/modules/iam_role_policy/self_managed_credentials/policy.json
+++ b/terraform/modules/iam_role_policy/self_managed_credentials/policy.json
@@ -64,6 +64,7 @@
           "Sid": "DenyAllExceptListedIfNoMFA",
           "Effect": "Deny",
           "NotAction": [
+              "iam:ChangePassword",
               "iam:CreateVirtualMFADevice",
               "iam:EnableMFADevice",
               "iam:GetUser",


pburkholder:
@ccostino What do you think?

ccostino:
Oh, yes, this would have to be the case upon account creation and setup.  When you first login into AWS with the accounts we create for folks who are onboarding, you are forced to change your password (which is the right step and a part of our compliance requirements as is), and in order to setup MFA you need to be able to login to the system so that you can attach a device/configure MFA.

However, once MFA is setup, I don't think we'd want this to continue to be the case, but I am not sure how you could account for that.  I don't know if there's a notion within a policy definition that states that this is allowed only upon first login or something to that effect.

pburkholder:
I need to bear in mind that the intent overall is enforce use of short-lived keys, so I don't think this change introduces anything that is a step back in our posture.

ccostino:
That's true!  But changing one's password is definitely an action that we'd only want the authorized user to perform.

That said, I think I'm overthinking this, too.  If MFA is enabled on an account, then you're going to need to provide the token to perform the action one way or another anyway, correct?  Regardless if it's through the web console or the command line.

PR #752, opened 2020-06-10
add user and policy for external domain acceptance tests

bengerman13:
- add an acceptance tests user
- add a route53 zone for acceptance tests

## security considerations
None

### DIFF
--- a/terraform/stacks/external/outputs.tf
+++ b/terraform/stacks/external/outputs.tf
@@ -21,23 +21,23 @@ output "external_domain_broker_secret_access_key_curr" {
 
 /* external domain broker test user */
 output "external_domain_broker_tests_username" {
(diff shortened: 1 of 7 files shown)


terraform/stacks/external/outputs.tf
```diff
@@ -19,6 +19,27 @@ output "external_domain_broker_secret_access_key_curr" {
   value = "${module.external_domain_broker.secret_access_key_curr}"
 }
 
+/* external domain broker test user */
+output "external_domain_broker_tests_username" {
+  value = "${module.external_domain_broker.username}"
```
soutenniza, line 24: Is value supposed to point to this?

bengerman13: :doh:

PR #755, opened 2020-06-17
add external domain broker user for govcloud partition

bengerman13:
- create an exernal-domain-broker user with correct permissions in govcloud

## security considerations
Adds a new user with permissions.

### DIFF
--- a/terraform/modules/external_domain_broker_govcloud/resources.tf
+++ b/terraform/modules/external_domain_broker_govcloud/resources.tf
@@ -12,7 +12,7 @@ data "template_file" "policy" {
 }
 
 resource "aws_iam_user" "iam_user" {
-  name = "external-domain-broker-${var.stack_description}"
+  name = "external-domain-broker-gov-${var.stack_description}"
 }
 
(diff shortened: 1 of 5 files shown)


terraform/modules/external_domain_broker_govcloud/resources.tf
```diff
@@ -0,0 +1,26 @@
...
+  template = "${file("${path.module}/policy.json")}"
+
+  vars {
+    account_id        = "${var.account_id}"
+    stack             = "${var.stack_description}"
+  }
+}
+
+resource "aws_iam_user" "iam_user" {
+  name = "external-domain-broker-${var.stack_description}"
```
bengerman13, line 15: worth noting that this user has the same name as the user in commercial. This should not create any direct conflict, but maybe it's confusing?

soutenniza: I would say that adding `-govcloud` would make it easier for future selfs. Everything else looks good

bengerman13: added `-gov`

PR #776, opened 2020-08-11
allow aws broker to access all parameter groups

bengerman13:
- Allow aws_broker to read all db parameter groups. This is necessary because the broker iterates over the parameter groups, and the iteration can't be filtered.

fixes cloud-gov/product#1405

## security considerations
Gives this broker more read permissions, but not much, and it's necessary for the operation of the broker

### DIFF
--- a/terraform/modules/iam_role_policy/aws_broker/policy.json
+++ b/terraform/modules/iam_role_policy/aws_broker/policy.json
@@ -14,7 +14,6 @@
         "rds:CreateDBParameterGroup",
         "rds:ModifyDBParameterGroup",
         "rds:DeleteDBParameterGroup",
-        "rds:DescribeDBParameterGroups",
         "rds:DescribeDBParameters",
         "rds:DescribeDBSnapshots",
         "rds:DeleteDBSnapshot"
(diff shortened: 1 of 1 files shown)


terraform/modules/iam_role_policy/aws_broker/policy.json
```diff
@@ -26,6 +25,15 @@
         "arn:${aws_partition}:rds:${aws_default_region}:${account_id}:subgrp:${rds_subgroup}"
       ]
     },
+    {
+      "Effect": "Allow",
+      "Action": [
+        "rds:DescribeDBParameterGroups"
+      ],
+      "Resource": [
+        "arn:${aws_partition}:rds:${aws_default_region}:${account_id}:pg:*"
```
ccostino, line 34: Just double checking, which resource is this that's being granted permissions for this action?  Also, removing the `rds:DescribeDBParameterGroups` from the other resource won't cause any other issues?

bengerman13: This is granting `rds:DescribeDBParameterGroups` on all parametergroups in our account

ccostino: Thanks!

PR #777, opened 2020-08-19
Elasticsearch Infrastructure

soutenniza:
- Subnets so Elasticsearch can have its own set of IPs to use
- Security Group which will only allow https (443) to elasticsearch endpoint
-

## security considerations
Makes Elasticsearch even more secure

### DIFF
--- a/terraform/modules/elasticsearch_broker/security_group.tf
+++ b/terraform/modules/elasticsearch_broker/security_group.tf
@@ -1,0 +1,23 @@
+resource "aws_security_group" "elasticsearch" {
+  description = "Allow access to incoming elasticsearch traffic"
+  vpc_id = "${var.vpc_id}"
+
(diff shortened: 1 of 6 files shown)


terraform/modules/elasticsearch_broker/security_group.tf
```diff
@@ -0,0 +1,23 @@
...
+    to_port = 443
+    protocol = "tcp"
+    security_groups = ["${var.security_groups}"]
+  }
+
+  egress {
+    from_port = 0
+    to_port = 0
+    protocol = "-1"
+    security_groups = ["${var.security_groups}"]
```
bengerman13, line 16: if `${var.security_groups}` is a `list`, do we need the braces here, or will that resolve to `[[group, other-group],]`?

PR #780, opened 2020-08-25
Update Terraform-managed RDS instances

ccostino:
This changeset is a multi-part series of steps to upgrade our Terraform-managed RDS instances that are internal to the cloud.gov platform.  When all steps are completed we should see the following:

- Change module names to not reference a specific PostgreSQL version

## Security considerations

## Maintenance considerations

## Review requests

- Is this the right approach to perform these modifications; do these in phases?  Should these be separate PRs in that case?
- Did I miss anything?
- One of the files has a parameter group explicitly defined, but it is the default parameter group; is this necessary?

### DIFF
--- a/terraform/modules/rds/variables.tf
+++ b/terraform/modules/rds/variables.tf
@@ -23,7 +23,7 @@ variable "rds_db_engine" {
 }
 
 variable "rds_db_engine_version" {
-  default = "9.6.15"
+  default = "11.1"
 }
 
 variable "rds_username" {}
@@ -45,7 +45,7 @@ variable "rds_parameter_group_name" {
 }
 
 variable "rds_parameter_group_family" {
-  default = "postgres9.6"
+  default = "default.postgres11"
 }
 
 variable "rds_multi_az" {
--- a/terraform/modules/cloudfoundry/database.tf
+++ b/terraform/modules/cloudfoundry/database.tf
@@ -1,4 +1,4 @@
-module "cf_database_96" {
+module "cf_database" {
   source = "../rds"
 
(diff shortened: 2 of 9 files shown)


terraform/modules/rds/variables.tf
```diff
@@ -45,7 +45,7 @@ variable "rds_parameter_group_name" {
 }
 
 variable "rds_parameter_group_family" {
-  default = "postgres9.6"
+  default = "default.postgres11"
```
ccostino, line 48: Is this even necessary?  This was just referring to the default parameter group, can we just remove this all together and let Terraform and AWS go with the default themselves?

bengerman13:
I'm a little concerned about having all this in one PR - that means that all these DBs will go offline for their updates at the same time, doesn't it?

ccostino:
🤔 yeah it would; is there a better area to focus first or just pick and choose one to start with?

ccostino:
I think am :overthinking: this; I will open separate PRs for each module to be merged on their own.

ccostino:
Closing this PR in favor of separate PRs:

- [cloudfoundry](https://github.com/cloud-gov/cg-provision/pull/781)
- [concourse](https://github.com/cloud-gov/cg-provision/pull/782)
- [rds module](https://github.com/cloud-gov/cg-provision/pull/783)
- [bosh](https://github.com/cloud-gov/cg-provision/pull/784)
- [credhub](https://github.com/cloud-gov/cg-provision/pull/785)
- [bosh (spoke)](https://github.com/cloud-gov/cg-provision/pull/786)

PR #783, opened 2020-08-28
Update Terraform-managed RDS instances - rds module

ccostino:
This changeset is part of a multi-step series of actions to upgrade our Terraform-managed RDS instances that are internal to the cloud.gov platform.  When all steps are completed we should see the following:

- Update default parameter group to `default.postgres11`

## Security considerations

## Maintenance considerations


terraform/modules/rds/variables.tf
```diff
@@ -45,7 +45,7 @@ variable "rds_parameter_group_name" {
 }
 
 variable "rds_parameter_group_family" {
-  default = "postgres9.6"
+  default = "default.postgres11"
```
ccostino, line 48: Is this even necessary? This was just referring to the default parameter group, can we just remove this all together and let Terraform and AWS go with the default themselves?

jfredrickson5:
AWS RDS now allows upgrading from `9.6.19` to `12.4` directly. I updated the changes in this PR so that it now targets engine version `12.4` and parameter group family `postgres12` instead of `11.1` and `postgres11` respectively.

bengerman13:
Since this one updates the default, it should probably be the last to be merged, right?

ccostino:
I'm not sure of the proper order to merge all of these PRs, but if this is the case, yes on this one.  @ChrisMcGowan do you hav a sense of what order all of these should go in once they're ready?

jfredrickson5:
We can close this PR since I ended up having to make this particular change to implement another RDS instance upgrade.

PR #785, opened 2020-08-28
Update Terraform-managed RDS instances - credhub

ccostino:
This changeset is part of a multi-step series of actions to upgrade our Terraform-managed RDS instances that are internal to the cloud.gov platform.  When all steps are completed we should see the following:

## Security considerations

## Maintenance considerations


jfredrickson5:
AWS RDS now allows upgrading from `9.6.19` to `12.4` directly. I updated the changes in this PR so that it now targets engine version `12.4` instead of `11.1`.

terraform/modules/credhub/variables.tf
```diff
@@ -17,6 +17,10 @@ variable "rds_parameter_group_name" {
   default = ""
 }
 
+variable "rds_parameter_group_family" {
+  default = "postgres9.6"
```
ccostino, line 21: Should we update our default to match version 12 now given all of the updates?  Or is this remaining because this DB hasn't been touched yet and needs to be handled separately?

ccostino (commented):
Thanks, @jfredrickson5!  I had just one question (and can't approve, since I opened the PR). :-)

terraform/modules/credhub/variables.tf
jfredrickson5: There's one more DB to go that looks like it should be a separate upgrade, but yes, we'll want to update the default to 12.4 after all updates have been applied.

ccostino: Totally fine, which is why we had all of these split in the first place!  I just wanted to make sure I remembered correctly and we weren't missing something else. :-)  Thanks, @jfredrickson5!

bengerman13 (approved):

PR #787, opened 2020-09-04
Compliance should check/refresh limit stuff

pburkholder:

- Add permissions for compliance role to effectively use TrustedAdvisor and check quotas
-

## security considerations

Get, List are within compliance role boundaries


ccostino:
@pburkholder it looks like this PR already had its changes merged in via other means, or is it still not ready to go?

pburkholder:
I guess we're good then. Thanks for taking note.

PR #794, opened 2020-12-21
allow ingestor profile to read objects

bengerman13:
- Allow ingestors to get objects so they can reingest

## security considerations
adds permissions to role

We should revert this once that work is done


soutenniza (approved):
<3

PR #806, opened 2021-02-10
Pipeline synced up to use terraform 0.12.30

soutenniza:
- All environments are using the terraform 0.12.30 bin
- Can't use main bin because we have no switched all our terraforms to use 0.12.30 yet
- Removed nested lists

## security considerations
Nah


spgreenberg (approved):
LGTM

PR #807, opened 2021-02-12
add standalone credhub

spgreenberg:

Support the deployment of a standalone in credhub instance in staging and production. Credhub needs:

- DNS entries
- ALB routing rules on tooling-main
- A Postgres DB

_I followed the pattern of concourse as the end deployment should look similar with the same requirements. However, I noticed there are multiple places with duplication. I followed the pattern as I am not 100% sure exactly which parts are required._

## security considerations

The credhub deployment will be similar to concourse. The VPN is required to access it as the tooling-main alb is used.

### DIFF
--- a/terraform/modules/stack/base/base.tf
+++ b/terraform/modules/stack/base/base.tf
@@ -15,6 +15,7 @@ module "vpc" {
   nat_gateway_instance_type         = var.nat_gateway_instance_type
   monitoring_security_groups        = var.target_monitoring_security_groups
   concourse_security_groups         = var.target_concourse_security_groups
+  credhub_security_groups           = var.target_credhub_security_groups
 }
 
 module "rds_network" {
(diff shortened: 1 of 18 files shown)


bengerman13 (dismissed):

terraform/modules/stack/base/base.tf
```diff
@@ -15,6 +15,7 @@ module "vpc" {
   nat_gateway_instance_type         = var.nat_gateway_instance_type
   monitoring_security_groups        = var.target_monitoring_security_groups
   concourse_security_groups         = var.target_concourse_security_groups
+  credhub_security_groups           = var.target_credhub_security_groups
```
bengerman13, line 18: the terraform check is complaining about this. I think you need to update the vpc module to know about this variable?

spgreenberg: It looks like I missed this: https://github.com/cloud-gov/cg-provision/pull/807/commits/e9eb960a25e05ee18aec397de90dfa3f2f7b864e#diff-6b3bfbeef0b5f97151e8df46b14b1ac7913ad73d151fe8afe19c60c5989f7440R114

There is also this concourse reference but it is in k8s module. I can't imagine this actually needs to be there so I am not adding credhub to it. https://github.com/cloud-gov/cg-provision/blob/39c50d95dfb77ec0e13f00d28f1fb944badaa21a/terraform/stacks/main/stack.tf#L173

bengerman13 (approved):

PR #808, opened 2021-02-18
Add permission to aws-broker rds role policy

ccostino:
This changeset adds a single permission, rds:DescribeDBEngineVersions, to the role policy for the aws-broker.  This will allow us to retrieve information necessary for generating new custom parameter groups that we would not be able to otherwise.

This is in support of https://github.com/cloud-gov/aws-broker/pull/148, which requires this change.

- Add Allow rule for `rds:DescribeDBEngineVersions`

## security considerations:
- This grants another read operation permission to the `aws-broker`, which already has other permissions to read information about the RDS environment and instances within our AWS account


soutenniza (approved):
+1

PR #829, opened 2021-05-03
Removed terratest because terraform bin is now 0.14.11

soutenniza:
- Terratest is removed so default "terraform" is used.

## security considerations
Nope

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -44,7 +44,6 @@ jobs:
       inputs:
       - name: pull-request
       params:
-        TERRAFORM_BIN: terratest-14.11
       run:
         path: pull-request/validate.sh
     on_success:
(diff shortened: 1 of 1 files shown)


ci/pipeline.yml
```diff
@@ -44,7 +44,6 @@ jobs:
       inputs:
       - name: pull-request
       params:
-        TERRAFORM_BIN: terratest-14.11
```
ccostino, line 47: Does this `params` section need to stay now that there's nothing in it?

ccostino (commented):
Thanks, @soutenniza!  One quick question.

ci/pipeline.yml
soutenniza: Better to leave it there in case we need it again

ccostino (approved):
Thanks again, @soutenniza!

PR #844, opened 2021-06-16
Bump all RDS to 12.5

pburkholder:
The `rds_apply_immediately` values are all `false` so this should only
take effect during the next mainteance window.

- Bumps version of RDS to 12.5
-
-

## security considerations

No information leaked in PR that's not already public. Applying this improves our compliance and security posture.


ccostino (approved):
Thanks, @pburkholder!

PR #855, opened 2021-07-19
Add route53 records for *.pages-staging.cloud.gov and *.sites.pages-staging.cloud.gov

davemcorwin:
- Adds wildcard DNS record/certs for
  - `*.pages-staging.cloud.gov`
  - `*.sites.pages-staging.cloud.gov`

My understanding is that this will allow us to create the following domains in CF via `create-domain`:
- `pages-staging.cloud.gov`
- `sites.pages-staging.cloud.gov`

enabling the creation of routes in CF via `create-route` such as:
- `app.pages-staging.cloud.gov`
- `admin.pages-staging.cloud.gov`
- `*.sites.pages-staging.cloud.gov`
- ...

## security considerations
???

### DIFF
--- a/terraform/stacks/dns/stack.tf
+++ b/terraform/stacks/dns/stack.tf
@@ -1060,6 +1060,30 @@ resource "aws_route53_record" "cloud_gov_star_fr_cloud_gov_aaaa" {
   }
 }
 
+resource "aws_route53_record" "cloud_gov_star_fr_cloud_gov_a" {
+  zone_id = aws_route53_zone.cloud_gov_zone.zone_id
+  name    = "*.sites.pages-staging.cloud.gov."
+  type    = "A"
+
+  alias {
+    name                   = "dualstack.${data.terraform_remote_state.production.outputs.cf_lb_dns_name}"
+    zone_id                = var.cloudfront_zone_id
+    evaluate_target_health = false
+  }
+}
+
+resource "aws_route53_record" "cloud_gov_star_fr_cloud_gov_aaaa" {
+  zone_id = aws_route53_zone.cloud_gov_zone.zone_id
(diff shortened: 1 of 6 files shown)


terraform/stacks/dns/stack.tf
```diff
@@ -1036,6 +1036,54 @@ resource "aws_route53_record" "dev2_delegate" {
   ]
 }
 
+resource "aws_route53_record" "cloud_gov_star_fr_cloud_gov_a" {
```
bengerman13, line 1039: these should all have unique values for the second string, shouldn't they? (e.g `cloud_gov_star_fr_cloud_gov_a`)

davemcorwin: absolutely, my mistake, fix has been pushed.

bengerman13:
This still needs some work - capturing notes here so we don't forget what it is:
- add certificate job to ci/pipeline.yml to fetch `*.pages-staging.cloud.gov` and `*.sites.pages-staging.cloud.gov` certificates. These should be copies of the `*.app.cloud.gov` version of the job, basically.
- add data blocks in terraform/stacks/main/stack.tf to fetch the data about the new certs (basically copies of these:  https://github.com/cloud-gov/cg-provision/blob/9810c967791c821c15bf07d279f80188ac503060/terraform/stacks/main/stack.tf#L30-L38)
- modify terraform/modules/cloudfoundry/elb_apps.tf and terraform/modules/cloudfoundry/elb_apps.tf to allow (require?) adding these certs. We'll need this module to do it https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb_listener_certificate
- change the target to `[...].cf_apps_lb_dns_name`

bengerman13 (approved):
LGTM

PR #886, opened 2021-10-05
moving to 25% http 75% https

ChrisMcGowan:
- Shift to 25% http and 75% https traffic in secureproxy

## security considerations
By removing http from secureproxy we are removing any non-tls from the path of elb to container


bengerman13 (approved):

soutenniza (approved):
<3

PR #896, opened 2021-11-01
Uaa elb creation with base waf rule

ChrisMcGowan:
- Create ELBs in each environment
- Create new target groups for elb in each environment
- Create new webacl for WAF with default AWS Managaed Core Rules for each environment
-  Set dev elb DNS records

## security considerations
This is the start of moving uaa to behind it's own elb with WAF so we can limit uaa exposure and overtime break out the customer auth path from app path


soutenniza (approved):
+1

PR #897, opened 2021-11-03
dns records for uaa elb for stage

ChrisMcGowan:
- Add dns records to move uaa/login.fr-stage.cloud.gov to point to uaa elb

## security considerations
Moving to uaa elb with WAF allow us to better secure the uaa/login path of the platform


soutenniza (approved):
+1, a few fixes was done via Google Meets

PR #909, opened 2021-11-19
require all child checks to be healthy

bengerman13:
- require all child checks to be healthy. I had the threshold reversed initially: the `1` before meant `consider this healthy if at least 1 child is healthy`, but what we want is for all the children to be healthy

## security considerations
None


mheadd (approved):
I also want all the children to be healthy

PR #916, opened 2021-12-03
output user data so we can use it

bengerman13:
- output user data so we can use it

## security considerations
None


davemcorwin (approved):
👍

PR #923, opened 2021-12-17
Jndi regex

soutenniza:
- This adds Regex Pattern for JDNI
- Added usage of AWS managed ruleset for log4j
- Adds the WAF rules to uaa web acl to block
- Associates all the LBs to that web acl so we can remove them later slowly

## security considerations
This is to match what's applied manually


ccostino (approved):
Thanks, @soutenniza!

PR #929, opened 2021-12-29
Let lets encrypt user assume roles

bengerman13:
- Let lets encrypt user assume roles

## security considerations
lets user have more access


rbogle (approved):
Ben and I discussed several methods for narrowing the scope of roles assumable by this policy change. And came up with using a named pattern. Should be relatively safe and easy to adhere to.

PR #946, opened 2022-01-09
include fr in tooling-west domains

bengerman13:
- include fr in tooling-west domains

## security considerations
None


mheadd (approved):
🚀

PR #965, opened 2022-01-28
start moving dev to its own DNS setup

bengerman13:
this is phase one of the move - we're creating the new zone and adding the appropriate records to it.

once this is merged and deployed, we manually check the records. If everything looks good, we uncomment the ns record, wait for the TTL to expire, then revalidate. 
If that's good, we delete the whole dev_old.tf file, and re-re-validate, and move on to staging.

- start moving dev to its own DNS setup

## security considerations
None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,7 @@ tmp/*
 # Terraform files
 *.tfstate
 *.tfstate.backup
-.terraform
+.terraform*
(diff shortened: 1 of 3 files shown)


apburnes (approved):

ccostino (approved):
Thanks, @bengerman13!

PR #985, opened 2022-02-09
FIPS 140: Add govcloud us-west fips endpoints

pburkholder:
-
- Where we have to chose between fips and non-fips endpoints, I've added the FIPS endpoints
- For many services, one no longer has to choose between FIPS and NON-FIPS
- I've only listed the services I think we're using, or likely to use.
- This is a draft while we figure out how to use with us-east, and whether this needs to be handled differently.

## security considerations
FIPS endpoints are already public. This is a compliance requirement, has nothing to do with security.


pburkholder:
@cloud-gov/platform-ops Thoughts on this?

terraform/stacks/main/stack.tf
```diff
@@ -3,7 +3,36 @@ terraform {
...
 provider "aws" {
+  # endpoints from # These endpoints are form https://docs.aws.amazon.com/govcloud-us/latest/UserGuide/using-govcloud-endpoints.html
+  # with help from https://github.com/Kaydub00/terraform-aws-fips/blob/master/govcloud-west.tf
+  endpoints {
+    # cloudtrail: Non-fips is n/a
+    # cloudwatch: non-fips is n/a
+    # cloudwatchevents: non-fips is n/a
+    # cloudwatchlogs: non-fips is n/a
+    # ec2             : non-fips is n/a
+    efs = "https://elasticfilesystem-fips.us-gov-west-1.amazonaws.com"
```
bengerman13, line 16: Can we generalize these so they work with regions other than us-gov-west-1 and partitions other than us-gov?

pburkholder: Not sure how. The EFS endpoints are:
* US E/W
  elasticfilesystem-fips.us-east-1.amazonaws.com
  elasticfilesystem-fips.us-east-2.amazonaws.com
  elasticfilesystem-fips.us-west-1.amazonaws.com
  elasticfilesystem-fips.us-west-2.amazonaws.com
* US Govcloud
  elasticfilesystem-fips.us-gov-east-1.amazonaws.com
  elasticfilesystem-fips.us-gov-west-1.amazonaws.com

According to https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/custom-service-endpoints:
> If multiple, different Terraform AWS Provider configurations are required, see the [Terraform documentation on multiple provider instances](https://www.terraform.io/docs/configuration/providers.html#alias-multiple-provider-instances) for additional information about the alias provider configuration and its usage.

bengerman13: I think we could parametrize this with, e.g. `http://elasticfilesystem-fips.${var.aws_region}.amazonaws.com`, but we'll have to do it for each provider (not a huge deal, just tedious and probably not super-obvious in some places)

terraform/stacks/main/stack.tf
```diff
@@ -3,7 +3,50 @@ terraform {
...
+  region = "us-gov-east-1"
+    endpoints {
+    efs = "https://elasticfilesystem-fips.us-gov-east-1.amazonaws.com"
+    es = "https://es-fips.us-gov-east-1.amazonaws.com"
+    firehose = "https://firehose-fips.us-gov-east-1.amazonaws.com"
+    kms = "https://kms-fips.us-gov-east-1.amazonaws.com"
+    lambda = "https://lambda-fips.us-gov-east-1.amazonaws.com"
+    wafv2 = "https://wafv2-fips.us-gov-east-1.amazonaws.com"
+  }
 }
```
bengerman13, line 50: I don't think this will work how you intend - when we use a different region, we use the same base provider and use environment variables so the provider is configured for the other region

pburkholder:
Last try for the day....

bengerman13:
I _think_ this will work as-is (after resolving conflicts), but will also need to be duplicated into other stacks

pburkholder:
I'd like to prioritize this tweak for at least partial implementation soon, so we can demonstrate some progress on our FIPS-140 adoption. Any takers?

rbogle:
@pburkholder TF check is failing. It looks like the $(var.aws_region} should be ${var.aws_default_region} in the providers. 
Ben indicated this needs to be duplicated in the rest of the stacks as well. 🤷‍♂️

pburkholder:
@bengerman13 `${var.aws_default_region}` or `${var.aws_region}` ?

bengerman13:
`${var.aws_default_region}`

pburkholder:
These checks have passed, good to go?

rcgottlieb (approved):

pburkholder:
Merging. The PR should change nothing since I don't think our `main` stack uses any of these APIs, but at least we're covered for when we do.

PR #994, opened 2022-02-11
add port specification, scale listeners more

bengerman13:
Context: turns out with elbv2, there's a 1:1 port:listener mapping, so we need to specify which port a listener listens on and add more listeners if we want more ports.
There's a limit of 50 ports per listener that we'll need to look out for.

- specify port
- scale listeners so we can have more ports

## security considerations
None


ccostino (approved):
Thanks, @bengerman13!

PR #1003, opened 2022-02-16
emit ports for tcp lbs

bengerman13:
- emit ports for tcp lbs

## security considerations
None


mheadd (approved):
👍

PR #1008, opened 2022-02-17
add tcp load balancer aliases

bengerman13:
- add tcp load balancer aliases

## security considerations
None


ccostino (approved):
Thanks, @bengerman13!

PR #1011, opened 2022-02-22
Add dev env, add commented out prod env, add dns for main subdomain for all envs

davemcorwin:
- DNS for development pages environment
- Stubs DNS for production pages environment
- DNS for main subdomain (ex pages.cloud.gov) for all envs

The diff is a little messy but the only domain being added for pages-staging is `pages-staging.cloud.gov`. All "dev" domains are new as well as the commented out ones for "prod".

## security considerations
Additional DNS and certificates for pages environments

## Questions
1. Can the DNS for pages(-env).cloud.gov be a CNAME? The wildcards ones are A/AAAA

### DIFF
--- a/terraform/modules/cloudfoundry/elb_apps.tf
+++ b/terraform/modules/cloudfoundry/elb_apps.tf
@@ -51,51 +51,16 @@ resource "aws_lb_listener" "cf_apps_http" {
   }
 }
 
-
-
-# resource "aws_lb_listener_certificate" "main_pages" {
-#   listener_arn    = aws_lb_listener.cf_apps.arn
-#   certificate_arn = var.main_pages_cert_id
-# }
-
-# resource "aws_lb_listener_certificate" "pages" {
-#   listener_arn    = aws_lb_listener.cf_apps.arn
-#   certificate_arn = var.pages_cert_id
-# }
-
-# resource "aws_lb_listener_certificate" "sites_pages" {
-#   listener_arn    = aws_lb_listener.cf_apps.arn
-#   certificate_arn = var.sites_pages_cert_id
-# }
-
-resource "aws_lb_listener_certificate" "main_pages_staging" {
+resource "aws_lb_listener_certificate" "pages" {
+  for_each        = var.pages_cert_ids
   listener_arn    = aws_lb_listener.cf_apps.arn
-  certificate_arn = var.main_pages_staging_cert_id
+  certificate_arn = each.key
 }
 
-resource "aws_lb_listener_certificate" "pages_staging" {
+resource "aws_lb_listener_certificate" "pages_wildcard" {
+  for_each        = var.pages_wildcard_cert_ids
   listener_arn    = aws_lb_listener.cf_apps.arn
-  certificate_arn = var.pages_staging_cert_id
-}
-
-resource "aws_lb_listener_certificate" "sites_pages_staging" {
-  listener_arn    = aws_lb_listener.cf_apps.arn
-  certificate_arn = var.sites_pages_staging_cert_id
-}
-
-resource "aws_lb_listener_certificate" "main_pages_dev" {
-  listener_arn    = aws_lb_listener.cf_apps.arn
-  certificate_arn = var.main_pages_dev_cert_id
-}
-
-resource "aws_lb_listener_certificate" "pages_dev" {
-  listener_arn    = aws_lb_listener.cf_apps.arn
-  certificate_arn = var.pages_dev_cert_id
-}
-
-resource "aws_lb_listener_certificate" "sites_pages_dev" {
-  listener_arn    = aws_lb_listener.cf_apps.arn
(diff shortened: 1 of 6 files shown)


terraform/modules/cloudfoundry/elb_apps.tf
```diff
@@ -63,6 +83,21 @@ resource "aws_lb_listener_certificate" "sites_pages_staging" {
   certificate_arn = var.sites_pages_staging_cert_id
 }
 
+resource "aws_lb_listener_certificate" "main_pages_dev" {
```
bengerman13, line 86: maybe a big side-track from your current goals, but can we put all of this into a module so we don't have to think about it in every environment?

davemcorwin: I thought about it, but i don't have a good enough handle on the organization of this repo to do it and this will be the last environment so 🤷 . I'm happy either way.

bengerman13:
Could we make this more modular:
- use a single list of certificate patterns as an input, instead of several fixed names. This would mean we can pass an empty list in environments other than production
- use a hosted zone for each pages dns domain, and use a single module to control all of them. This makes it easier to move your DNS around in the future if you ever want it to be in its own AWS account, for example. It also makes our DNS config a lot easier to navigate, both in TF and in the AWS console

davemcorwin:
> Could we make this more modular:
> 
>     * use a single list of certificate patterns as an input, instead of several fixed names. This would mean we can pass an empty list in environments other than production

Since we need to generate resource identifiers which can't include `.`, I think this would need to be a map of pattern and name OR we would have to apply some sort of transformation to the pattern to generate the identifier, i.e. [`replace`](https://www.terraform.io/language/functions/replace), which can get a little messy/error prone.

In the Concourse pipeline, this would need to be specified as one big string, which is a bit awkward AND it doesn't help us in the pipeline itself because we can't parameterize the jobs in the necessary way, correct? 



>     * use a hosted zone for each pages dns domain, and use a single module to control all of them. This makes it easier to move your DNS around in the future if you ever want it to be in its own AWS account, for example. It also makes our DNS config a lot easier to navigate, both in TF and in the AWS console

Do you mean a hosted zone for each environment? It looks like that's what CG has? I'm not that familiar with Route53

Are you thinking something similar to the `environment_dns` module?

bengerman13:
> Do you mean a hosted zone for each environment? It looks like that's what CG has? I'm not that familiar with Route53

yeah - a hosted zone for each environment

>Since we need to generate resource identifiers which can't include ., I think this would need to be a map of pattern and name OR we would have to apply some sort of transformation to the pattern to generate the identifier, i.e. [replace](https://www.terraform.io/language/functions/replace), which can get a little messy/error prone.

> In the Concourse pipeline, this would need to be specified as one big string, which is a bit awkward AND it doesn't help us in the pipeline itself because we can't parameterize the jobs in the necessary way, correct?

I think it would work fine to turn these into a module, then loop over the module with a count. Here's about what I have in mind:

```
# in terraform/modules/cloudfoundry/variables.tf
# and also terraform/stack/main.tf
variable "extra_cert_patterns" {
}
```
```
# in terraform/modules/cloudfoundry/elb_apps.tf
module "pages_cert" {
  for_each = to_set(var.extra_cert_patterns)
  source   = "../../extra_certs"
}
```

```
# in ci/pipeline.yml
  - name: plan-bootstrap-development
    plan:
    [...]
    params: &development-params
      [...]
      TF_VAR_extra_cert_patterns: '[]'
[...]
  - name: plan-bootstrap-staging
    plan:
    [...]
    params: &staging-params
      [...]
      TF_VAR_extra_cert_patterns: '[]'
 [...]
  - name: plan-bootstrap-production
    plan:
    [...]
    params: &production-params
      [...]
      TF_VAR_extra_cert_patterns: '["star.sites.pages-staging.cloud.gov","star.pages-staging.cloud.gov", [...]]'
```

bengerman13:
just realized - the dns change will also require some extra stuff to make cert renewals work

davemcorwin:
Which DNS change? Can we move forward as is or do we need to refactor?

bengerman13:
if we put the dns in its own hosted zone, it'll require a change to the cert renewal steps.

Let's leave the dns as-is, but make the change to use an array of cert patterns for the ALBs so we don't have to have extra certs in environments that don't otherwise need to think about Pages/Federalist

davemcorwin:
@bengerman13 This provides the cert patterns as an array.
1. this doesn't remove the need for individual jobs in the pipline for each cert, so they are still explicitly listed there.
2. I split out the wildcard cert arns because if we do decide to generate the dns from them, we can since the wildcards are handled differently than the others.

bengerman13 (approved):
looks great - Thanks Dave!

davemcorwin:
FYI - I'm not authorized to merge this.

PR #1013, opened 2022-02-23
Groups in wrong place

bengerman13:
- remove security group from NLB, since it doesn't go there
- create target group for each port, since apparently that's what AWS wants

## security considerations
allows more traffic into our VPC, but only where expected


ccostino (approved):
Thanks, @bengerman13!

PR #1014, opened 2022-02-24
point listeners to different targets

bengerman13:
- point listeners to different targets
-
-

## security considerations
None

### DIFF
--- a/terraform/stacks/main/outputs.tf
+++ b/terraform/stacks/main/outputs.tf
@@ -612,5 +612,5 @@ output "tcp_lb_listener_ports" {
 }
 
 output "tcp_lb_security_groups" {
-  value = [module.cf.tcp_lb_security_groups.*.name, module.stack.bosh_security_group]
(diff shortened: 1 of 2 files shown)


terraform/stacks/main/outputs.tf
```diff
@@ -612,5 +612,5 @@ output "tcp_lb_listener_ports" {
 }
 
 output "tcp_lb_security_groups" {
-  value = [module.cf.tcp_lb_security_groups.*, module.stack.bosh_security_group]
+  value = [module.cf.tcp_lb_security_groups.*.name, module.stack.bosh_security_group]
```
soutenniza, line 615: does the bosh_security_group need to reference to name too?

bengerman13: it's already a name

soutenniza (approved):
🥑

terraform/stacks/main/outputs.tf
bengerman13: I lied. It's an id

soutenniza (approved):
🐦

PR #1021, opened 2022-03-02
start moving DNS to environments

bengerman13:
- start moving DNS to environments. 

## Context

This is part of decoupling AWS accounts. I'm iterating on getting commercial pieces into their own accounts. 
This is also part of a plan to get each environment into a single Terraform stack. Next up in this plan is:
1.  moving DNS into the main stack
2. move external into main 
3. replace lets encrypt scripts with the lets encrypt terraform module

** after merging this, manual terraform state updates are required **

## security considerations
since this is part of AWS account reorg, it helps us isolate failure and compromise domains, as well as decreasing time to recovery in case of total environment loss


markdboyd:
This PR is very stale. We can re-open or revisit if/when this work resumes.

PR #1036, opened 2022-03-22
Ecr

soutenniza:
- ECR in tooling stack

## security considerations
User will only have access to ecr.

### DIFF
--- a/terraform/modules/iam_user/ecr_user/policy.json
+++ b/terraform/modules/iam_user/ecr_user/policy.json
@@ -1,0 +1,12 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Action": [
+        "ecr:*"
+      ],
+      "Effect": "Allow",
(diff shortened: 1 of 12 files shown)


apburnes (dismissed):

terraform/modules/iam_user/ecr_user/policy.json
```diff
@@ -0,0 +1,12 @@
+{
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Action": [
+        "ecr:*"
```
apburnes, line 6: Do we want to limit these permissions?

soutenniza: I'll limit it when I learn more about permissions and such :)

apburnes: 👍

apburnes (approved):

PR #1044, opened 2022-04-19
removing hard coded minor versions

soutenniza:
- Its so we dont revert to old versions

## security considerations
None


rbogle (approved):
👍

PR #1083, opened 2022-07-08
update s3 endpoint policy

rbogle:
- Augments existing policy for each S3 Gateway Endpoint to limit to account principals and resources. 
- provides CIDR ranges for s3 in output for cg-deploy-cf to use for trusted_networs_egress ASGs
- deploys a private endpoint to s3 for clients to use in the public ASG to get to us-gov-west-1 s3 buckets outside of our account
-  Outputs those ips addrs for the private endpoint so cg-deploy-cf and augment the public-ASG 

## security considerations
New rule ensures that access via the existing s3 private endpoint is restricted to principals in the account and resources in the account. Access to specific resources is controlled by bucket or IAM policy. For users to access s3 buckets not brokered by the platform, they will need to use the Public ASG and in the case of needing to access a bucket in a different account but in the same region as the platform, will need to use the private endpoint deployed here. 

resolves cloud-gov/product/issues/1775

### DIFF
--- a/terraform/modules/bosh_vpc/endpoint_s3.tf
+++ b/terraform/modules/bosh_vpc/endpoint_s3.tf
@@ -11,9 +11,20 @@ resource "aws_vpc_endpoint" "private-s3" {
         "Effect": "Allow",
         "Resource": "*",
         "Principal": "*"
+        "Condition": {
+          "ForAllValues:StringEquals": {
+            "aws:PrincipalAccount": ${account_id},
+            "aws:ResourceAccount": ${account_id}"
+          }				
+        }        
     }]
 }
 EOF
 
 }
 
+data "aws_caller_identity" "current" {}
+
+locals {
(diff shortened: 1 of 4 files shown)


mogul (approved):
This looks fine. 

(I was wondering if we'll run into problems where customers running apps on different foundations won't be able to have each app refer to S3 buckets in the other foundation. But then I realized customers would just handle that the same way as they will handle bucket-outside-our-account UPS use-case with a single foundation now, going out over the internet rather than through the gateway, and that's probably the right thing to do boundary-wise.)

rbogle:
Yes my intention would be that if using public-egress there would be no restrictions, and so the UPS case is supported. But still working through this. The public-egress valid UPS case breaks with the way this policy and ASG is written now, so Im working on some more testing.

mogul (approved):
Leaving for you to merge when appropriate.

PR #1092, opened 2022-07-18
Refactor TCP NLB code

markdboyd:

- Move code to create NLB, target groups, listeners, and security groups into a Terraform `tcp` module
- Refactor deployment code to invoke `tcp` module (conditionally) to create a TCP router for platform usage
- Allow custom tags to be specified and used when deploying all TCP NLB resources

## security considerations

None, just refactoring deployment code. But the additional facility for tagging will make resource usage more visible and trackable.


pburkholder:
I have no concerns but I'm not currently fluent enough in TF to approve

PR #1099, opened 2022-08-08
S3 egress access for trusted asgs

rbogle:
- Updates s3 gateway policy to limit to accounts we want and the ECR service bucket
-
-

## security considerations
None

### DIFF
--- a/terraform/modules/bosh_vpc/endpoint_s3.tf
+++ b/terraform/modules/bosh_vpc/endpoint_s3.tf
@@ -25,7 +25,9 @@ resource "aws_vpc_endpoint" "private-s3" {
       "Principal": "*",
       "Action": "s3:*",
       "Resource": [
-        "arn:aws:s3:::prod-us-gov-west-1-starport-layer-bucket/*"
+        "arn:aws:s3:::prod-us-gov-west-1-starport-layer-bucket/*",
+        "arn:aws:s3:::prod-us-gov-east-1-starport-layer-bucket/*",
+        "arn:aws:s3:::prod-us-east-1-starport-layer-bucket/*",
       ]
     }
     ]
--- a/terraform/modules/bosh_vpc/endpoint_s3.tf
+++ b/terraform/modules/bosh_vpc/endpoint_s3.tf
@@ -34,15 +34,6 @@ EOF
 
 }
 
-/* This needs to be in place for trusted s3 egress to work. Removing from policy ^^ due to ECR build issues with federalist
-        ,
-        "Condition": {
-          "ForAllValues:StringEquals": {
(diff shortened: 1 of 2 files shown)


terraform/modules/bosh_vpc/endpoint_s3.tf
```diff
@@ -11,22 +11,31 @@ resource "aws_vpc_endpoint" "private-s3" {
...
-        } 
-*/
+        }        
+    },
+    {
+      "Sid": "Access-to-ecr-buckets",
+      "Effect": "Allow",
+      "Principal": "*",
+      "Action": "s3:*",
+      "Resource": [
```
markdboyd, line 27: Probably an ignorant question: but do we not want the same conditions to restrict only traffic from the expected accounts to these buckets?

rbogle: I dont think it would be needed, this pol statement should allow the ECR service itself to access its own bucket outside our account. The principle account condition might not apply to the service principal in this case, or it might block the service. And by the nature of its deployment, this policy only applies to traffic within the account its deployed in, unless we use a shared or peered vpc across accounts...

The way the condition filter works in the top statement is to check that the Iam user/role making the request is one that lives in an account we own, and that the Resource also lives in that same set of accounts.. This would prevent a client from creating a user in their own account and using those credentials in a CF app to access an S3 bucket in their own account under trusted egress. For that they still need to use public_egress..

markdboyd: Sounds good. Though the `Principal: "*"` in this statement means it's not just ECR getting access to these buckets through the gateway, but in theory anything, correct?

markdboyd (approved):

terraform/modules/bosh_vpc/endpoint_s3.tf
rbogle: Looks like I might be able to use a condition with:
 "Condition": {
          "ForAllValues:StringEquals": {
            "aws:PrincipalServiceName": "ecr.amazonaws.com
    }
}

markdboyd: oK, just checking

rbogle: However even though the policy allows access to those buckets, unless the ECR service has those buckets open to the world, a user wouldn't be able to access them. They most likely have bucket policies in place..

rbogle: I appreciate the review and questions, Thanks Mark!

PR #1105, opened 2022-08-16
S3 ep policy

rbogle:
- adds a policy to s3 endpoints to limit traffic to s3 from known sources based on account ids. 
- includes a statement for ECR service accessing s3 
- outputs the cidrs and ips for use by deploy-cf to modify the ASGs.

## security considerations
None


rbogle:
closing becasue branch name is causing problems with pr test

PR #1113, opened 2022-08-29
feat: Update and add pages support domain for new zendesk

apburnes:
- Updates existing support.cloud.gov domain to the new Zendesk instance
- Add `pages-support.cloud.gov` domain for the new Zendesk instance team for cloud.gov Pages.

## security considerations
Needed for Zendesk migration

### DIFF
--- a/terraform/stacks/dns/validators.tf
+++ b/terraform/stacks/dns/validators.tf
@@ -30,7 +30,15 @@ resource "aws_route53_record" "cloud_gov_zendesk_support_cname" {
   name    = "support.cloud.gov."
   type    = "CNAME"
   ttl     = 60
-  records = ["cloud-gov.zendesk.com."]
+  records = ["cloud-gov-new.zendesk.com."]
+}
+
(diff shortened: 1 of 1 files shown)


terraform/stacks/dns/validators.tf
```diff
@@ -30,7 +30,15 @@ resource "aws_route53_record" "cloud_gov_zendesk_support_cname" {
   name    = "support.cloud.gov."
   type    = "CNAME"
   ttl     = 60
-  records = ["cloud-gov.zendesk.com."]
+  records = ["cloud-gov-new.zendesk.com."]
```
jameshochadel, line 33: Just to confirm, we don't need to preserve the old record?

apburnes: I believe they are shutting down the old subdomain `cloud-gov.zendesk.com` along with the old instance it is associated to as apart of the Zendesk migration.

jameshochadel (dismissed):

jameshochadel (approved):

PR #1118, opened 2022-09-13
add AWS managed rule for IP reputation to WAF rule set

markdboyd:

- add AWS managed rule for IP reputation to WAF rule set

## security considerations

This change should enhance security as it hardens our firewall rules to block traffic from known bad IPs


soutenniza (approved):
LGTM

PR #1121, opened 2022-09-15
Add log_encrypted_bucket module

dandersonsw:
- Adds a new module for s3 buckets and adds transition settings for the logging bucket

## security considerations
Helps meet log compliance requirements related to M-21-31

### DIFF
--- a/terraform/modules/s3_bucket/log_encrypted_bucket/log_encrypted_bucket.tf
+++ b/terraform/modules/s3_bucket/log_encrypted_bucket/log_encrypted_bucket.tf
@@ -32,6 +32,7 @@ resource "aws_s3_bucket_lifecycle_configuration" "log_encrypted_bucket_lifecycle
     filter {
       prefix  = ""
     }
+    #if expiration_days is 0 then the rule is disabled
     status = var.expiration_days == 0 ? "Disabled" : "Enabled"
     transition {
       days          = 365
--- a/terraform/modules/s3_bucket/log_encrypted_bucket/log_encrypted_bucket.tf
+++ b/terraform/modules/s3_bucket/log_encrypted_bucket/log_encrypted_bucket.tf
@@ -1,0 +1,71 @@
+resource "aws_s3_bucket" "log_encrypted_bucket" {
+  bucket        = var.bucket
+  force_destroy = var.force_destroy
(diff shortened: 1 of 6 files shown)


terraform/modules/s3_bucket/log_encrypted_bucket/log_encrypted_bucket.tf
```diff
@@ -0,0 +1,71 @@
...
+    status = var.expiration_days == 0 ? "Disabled" : "Enabled"
+    transition {
+      days          = 365
+      storage_class = "ONEZONE_IA"
+    }
+    expiration {
+      # Hack: Set expiration days to 30 if unset; objects won't actually be expired because the rule will be disabled
+      # See https://github.com/terraform-providers/terraform-provider-aws/issues/1402
+      days = var.expiration_days == 0 ? 30 : var.expiration_days
+    }
```
markdboyd, line 44: What actually disables the rule if it's set to expire in 30 days?

Also, did you see this comment in that thread? https://github.com/hashicorp/terraform-provider-aws/issues/1402#issuecomment-514480313

I think it might be preferable to use `for_each` to make the block truly conditional, so that we can guarantee that the expiration lifecycle rule is only added when we want.

I think you can use the `for_each` approach if you add this to the file somewhere:

```
locals {
  expiration_enabled = var.expiration_days == 0 ? [] : ["enabled"];
}
```

And then change this block like so:

```suggestion
  dynamic "expiration" {
      for_each = local.expiration_enabled
      content {
        days = var.expiration_days
      }
    }
```

dandersonsw: I just copied this code from the `encrypted_bucket` terraform code, and added the section for transition. I can try updating this with dynamic blocks, but right now the other buckets seem to be working.

dandersonsw: Added a comment showing where the rule gets disabled. I think adding an extra variable and a dynamic block would make this more difficult to understand and read.

markdboyd (approved):

markdboyd (approved):

PR #1131, opened 2022-09-28
Make encryption bucket policy optional

markdboyd:

- remove bucket policy requiring specific encryption for query resolver logging buckets

## security considerations

There shouldn't be any negative security impact here as once you have bucket encryption for S3 enabled, you can guarantee that there is some form of encryption. The bucket policy being made optional here just guarantees that all PUT requests to the bucket must use the specific form of encryption enabled on the bucket. [See this article for further explanation](https://aws-blog.com/2020/09/enforcing-encryption-standards-on-s3-objects.html).

While the bucket policy is valuable, it can't be enabled on buckets used for DNS resolver query logging because that results in "access denied" errors. And I created a bucket without this policy to prove that it could integrate successfully with Route53 resolver query logging. Since having this logging is valuable for security as well, it seems worthwhile to disable the bucket policy for these specific buckets.

### DIFF
--- a/terraform/modules/dns_logging/resolver_logging.tf
+++ b/terraform/modules/dns_logging/resolver_logging.tf
@@ -3,6 +3,9 @@ module "dns_resolver_logs_bucket" {
   bucket          = "${var.stack_description}-query-resolver-logs"
   aws_partition   = var.aws_partition
   expiration_days = 930 # 31 days * 30 months = 930 days
+  # need to exclude bucket policy otherwise we get access denied errors from rout53
+  # query resolver logging configuration
+  include_require_encrypted_put_bucket_policy = false
 }
 
 resource "aws_route53_resolver_query_log_config" "resolver_config" {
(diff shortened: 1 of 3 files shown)


rcgottlieb (approved):
It's not a big deal, but there's a typo in resolver_logging.tf line 6 there is rout53 instated of route53. I'm approving anyway.

terraform/modules/dns_logging/resolver_logging.tf
```diff
@@ -3,6 +3,9 @@ module "dns_resolver_logs_bucket" {
   bucket          = "${var.stack_description}-query-resolver-logs"
   aws_partition   = var.aws_partition
   expiration_days = 930 # 31 days * 30 months = 930 days
+  # need to exclude bucket policy otherwise we get access denied errors from Route53
+  # query resolver logging configuration
```
pburkholder, line 7: expand this comment to add:
```The bucket's default encryption setting is still enforced to ensure "encryption at rest"```

rcgottlieb (approved):

terraform/modules/dns_logging/resolver_logging.tf
markdboyd: aecef1a3c82eef3bf00de72856f702ee3aadc29c

rcgottlieb (approved):

PR #1135, opened 2022-10-04
add athena database for query resolver logs

markdboyd:

- add athena database for query resolver logs so we can easily search the query resolver logs in S3

## security considerations

None


jameshochadel:
I recall hearing that any change that would alter an architecture diagram requires an SCR. Does shipping logs to a new place change the architecture diagram?

markdboyd:
@jameshochadel Yeah I'm going to put this PR in draft because we're not sure if we're going to do this

PR #1140, opened 2022-10-07
update monitoring SG to restrict ingress traffic to VPC CIDR

markdboyd:

- update monitoring SG to restrict ingress traffic to VPC CIDR

## security considerations

This change enhances security by removing public access on all ports and limiting access to inbound traffic from the VPC

### DIFF
--- a/terraform/modules/monitoring/sg_monitoring.tf
+++ b/terraform/modules/monitoring/sg_monitoring.tf
@@ -19,7 +19,6 @@ resource "aws_security_group_rule" "self_reference" {
   from_port         = 0
   to_port           = 0
   protocol          = -1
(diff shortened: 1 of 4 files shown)


pburkholder:
I ❤️ this so much!

pburkholder:
How are the values for `var.vpc_cidr` being set and tested?

markdboyd:
@pburkholder Those values seem to come from the pipeline definition: https://github.com/cloud-gov/cg-provision/blob/main/ci/pipeline.yml#L306. So the values must be in CredHub or a YAML credentials file. 

I'm not sure what you mean by "tested" though

jameshochadel (approved):
🚀

PR #1141, opened 2022-10-13
Test CloudFront module by fronting one LB in staging

jameshochadel:

- Create two new modules to describe CloudFront resources:
  - `modules/cloudfront/distribution` configures a single distribution that fronts an LB, protected by Shield Advanced
  - `modules/cloudfront/environment` protects all LBs in an environment with distributions
- Create a new stack, `stacks/cloudfront`, that will instantiate the `cloudfront/environment` module for each environment

To test the change, CloudFront will only front one LB in staging, and DNS will not point to it for now.

## Security Considerations

This change only affects the staging environment and is intended to improve resiliency of the platform to DDoS attacks.


jameshochadel:
I am happy to report that this passed a `terraform plan` [here](https://ci.fr.cloud.gov/teams/main/pipelines/terraform-provision/jobs/plan-cloudfront/builds/10) and is ready for review.

terraform/modules/cloudfront/distribution/variables.tf
```diff
@@ -0,0 +1,20 @@
...
+  type = object({
+    lb_dns_name   = string
+    lb_http_port  = number
+    lb_https_port = number
+  })
+}
+
+# variable "iam_certificate_id" {
+#   type = string
+# }
```
markdboyd, line 11: Do we need this?

terraform/modules/cloudfront/distribution/distribution.tf
```diff
@@ -0,0 +1,56 @@
...
+
+  restrictions {
+    geo_restriction {
+      restriction_type = "none"
+    }
+  }
+
+  default_cache_behavior {
+    allowed_methods = ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
+    cached_methods  = ["GET", "HEAD"]
```
markdboyd, line 32: Do we want to have any caching on this Cloudfront? This distribution will be protecting all apps on *.app.cloud.gov, which can vary widely. If we implement caching, I fear that we might have difficulty with needing to clear the cache for some sites but not others. 

Would it be better to disable caching and just do a pass through to the load balancer?

ci/pipeline.yml
```diff
@@ -270,6 +271,33 @@ jobs:
       <<: *dns-params
       TERRAFORM_ACTION: apply
 
+- name: plan-cloudfront
```
markdboyd, line 274: So I think this setup is deploying to our GovCloud accounts, not commercial, which also seems to be indicated by the output in Concourse

terraform/stacks/cloudfront/staging.tf
```diff
@@ -0,0 +1,6 @@
+module "staging_cloudfront" {
```
markdboyd, line 1: Do we need a new `terraform/stacks/cloudfront`? We already have `terraform/stacks/external` which [manages deploying resources to our commercial accounts](https://github.com/cloud-gov/cg-provision/blob/main/README.md?plain=1#L38)

So we could just add this module to `terraform/stacks/external/stack.tf` and it will get auto-deployed to create dev, stage, and prod resources in our commercial accounts from pipeline jobs like [this one](https://github.com/cloud-gov/cg-provision/blob/main/ci/pipeline.yml#L77)

terraform/modules/cloudfront/distribution/variables.tf
jameshochadel: This is a placeholder for when I point the DNS to the distribution. I think each distribution will need to have the cert that its corresponding LB currently has.

terraform/modules/cloudfront/distribution/distribution.tf
jameshochadel: Per this & other discussions, I'm going to remove caching from the distribution 👍

ci/pipeline.yml
jameshochadel: Agreed — fixing.

terraform/stacks/cloudfront/staging.tf
jameshochadel: I'm strongly in favor of this idea, but I didn't address it here because of the cross-partition state dynamic we talked about a few days ago. I'd like to merge the external + dns + cloudfront stacks all together, but I'm hesitant to make a change to the external stack's state. It currently only talks to Commercial, I think — no reading GovCloud state.

I can look into it if you feel strongly though! One concern I have by having them separate is that Terraform won't be able to build its dependency graph properly and there will be ordering issues when something changes.

markdboyd:
@jameshochadel It all looks good to me except the changes in the README which are duplicated in #1163 . I see you had feedback on those changes in that PR, so is there any way to remove the README changes from this PR?

markdboyd (approved):
Approving, though the README changes should be updated with the feedback in #1163

PR #1151, opened 2022-10-24
add rule to rate limit non-CDN requests

markdboyd:

- add rule to rate limit non-CDN requests to 250,000 requests in 5 minutes

## security considerations

This change provides additional defense against DDoS attacks on our infrastructure.

### DIFF
--- a/terraform/modules/cloudfoundry/elb_uaa.tf
+++ b/terraform/modules/cloudfoundry/elb_uaa.tf
@@ -198,8 +198,32 @@ resource "aws_wafv2_web_acl" "cf_uaa_waf_core" {
   }
 
   rule {
-    name     = "CG-RegexPatternSets"
+    name = "RateLimitNonCDN"
     priority = 4
+
+    action {
+      block {}
+    }
+
+    statement {
+      rate_based_statement {
+        limit = 250000
+        aggregate_key_type = "FORWARDED_IP"
+        # Requests without forwarded IP headers will be counted towards the rate limit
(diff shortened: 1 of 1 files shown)


terraform/modules/cloudfoundry/elb_uaa.tf
```diff
@@ -198,8 +198,32 @@ resource "aws_wafv2_web_acl" "cf_uaa_waf_core" {
...
     priority = 4
+
+    action {
+      block {}
+    }
+
+    statement {
+      rate_based_statement {
+        limit = 250000
+        aggregate_key_type = "FORWARDED_IP"
```
jameshochadel, line 211: If this works based on the forwarded IP (e.g. `X-Forwarded-For`), won't it _only_ apply the CDN requests? Do other parts of our stack add that header?

markdboyd: I don't think so. This is the screenshot of the rule as currently configured on our dev WAF:

Screen Shot 2022-10-25 at 10 36 36 AM

So telling the WAF to do rate-limiting based on the forwarded IP means it won't rate limit based on the source IP, which might be the CDN IP address, but the IP address that sent the request to the CDN.

It's worth nothing that we may be okay to rate limit even CDN requests at 250,000 requests/5 minutes, but I'm just trying to be extra careful here.

jameshochadel (approved):
Ty sir! :shipit:

PR #1152, opened 2022-10-25
Enable deletion protection for ELBs #580. AWS Security improvements #…

dtamasang2:
…2272. This change was implemented to add deletion protection on all aws_lb resources in the Terraform code. These are the following files that were updated:

terraform/modules/
/admin - admin_elb_main.tf
/cloudfoundry - elf_apps.tf, elf_main.tf, elf_uaa.tf, nlb.tf /concourse - elf.tf
/diego - diego_elb_main.tf
/elesticache_broker_network - elb.tf
/logsearch - elb_platform_syslog.tf, elb.tf
/monitoring - elb.tf

terraform/stacks
/regionalmasterbosh - elb_uaa.tf
/tooling - elb_uaa.tf
-
-
-

## security considerations
This is a security change which makes it harder to delete elb resources.

### DIFF
--- a/terraform/modules/diego/diego_elb_main.tf
+++ b/terraform/modules/diego/diego_elb_main.tf
@@ -48,8 +48,6 @@ resource "aws_elb" "diego_elb_main" {
   subnets         = var.elb_subnets
   security_groups = [aws_security_group.diego-elb-sg.id]
 
-  enable_deletion_protection  = true
-
   listener {
     lb_port           = 2222
     lb_protocol       = "TCP"
--- a/terraform/modules/diego/diego_elb_main.tf
+++ b/terraform/modules/diego/diego_elb_main.tf
@@ -48,6 +48,8 @@ resource "aws_elb" "diego_elb_main" {
   subnets         = var.elb_subnets
   security_groups = [aws_security_group.diego-elb-sg.id]
 
+  enable_deletion_protection  = true
+
   listener {
     lb_port           = 2222
     lb_protocol       = "TCP"
--- a/terraform/stacks/regionalmasterbosh/stack.tf
+++ b/terraform/stacks/regionalmasterbosh/stack.tf
@@ -56,8 +56,9 @@ resource "aws_lb" "main" {
     bucket  = var.log_bucket_name
     prefix  = var.stack_description
     enabled = true
-  enable_deletion_protection  = true
   }
+ 
+  enable_deletion_protection  = true
 }
 
 resource "aws_lb_listener" "main" {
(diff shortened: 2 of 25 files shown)


terraform/modules/diego/diego_elb_main.tf
```diff
@@ -48,6 +48,8 @@ resource "aws_elb" "diego_elb_main" {
   subnets         = var.elb_subnets
   security_groups = [aws_security_group.diego-elb-sg.id]
 
+  enable_deletion_protection  = true
```
markdboyd, line 51: I don't think this property is supported on ELB resources?

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elb

markdboyd (changes requested):
I'm not sure that the ELB resources support an `enable_deletion_protection` property. Also, I think that we missed updating some of the `aws_lb` resources, such as the ones in [`terraform/stacks/main/domains_broker.tf`](https://github.com/cloud-gov/cg-provision/blob/main/terraform/stacks/main/domains_broker.tf#L12). We should do a code search for `aws_lb` resources to make sure we update all of them

terraform/modules/diego/diego_elb_main.tf
dtamasang2: I think It may still work since ultimately its still a load balancer type. Looking more into it.

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb

markdboyd: @dtamasang2 Actually, we have a Concourse job that runs to validate our Terraform. And as [you can see here](https://ci.fr.cloud.gov/builds/101844362), that job is failing with errors like:

```
│ Error: Unsupported argument
│ 
│   on ../../modules/elasticache_broker_network/elb.tf line 7, in resource "aws_elb" "elasticache_elb":
│    7:   enable_deletion_protection  = true
│ 
│ An argument named "enable_deletion_protection" is not expected here.
```

There is apparently a difference between what is supported by [`aws_elb` resources](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/elb), like the one in this error, and [`aws_lb` resources](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/lb).

terraform/stacks/regionalmasterbosh/stack.tf
```diff
@@ -56,6 +56,7 @@ resource "aws_lb" "main" {
     bucket  = var.log_bucket_name
     prefix  = var.stack_description
     enabled = true
+  enable_deletion_protection  = true
```
markdboyd, line 59: This line is in the wrong place. It needs to outside of this `access_logs` block, above current line 55

markdboyd (approved):

PR #1153, opened 2022-10-27
Add cloudwatch alarm

dandersonsw:
- Adds module for cloudwatch alarms
- Sets up alarm for increased 404s

## security considerations
This is a security improvement that will hopefully give us quicker notice when there's issues on the system


markdboyd:
@dandersonsw The PR looks fine, but can we talk about the cloudwatch alarm configuration before merging? I think we're seeing a lot of notifications from these alarms, so want to make sure that we've got the signal/noise ratio correct here

markdboyd (dismissed):

markdboyd (approved):

PR #1189, opened 2022-12-09
add missing sns_name var from stack

dandersonsw:
- Add sns_name variable to stack so terraform doesn't complain

## security considerations
None


markdboyd (dismissed):

dandersonsw:
This PR not actually needed

PR #1201, opened 2023-01-06
Update IAM policy for KMS encrypted bucket module

markdboyd:

- add support for specifying external accounts that should have bucket access
- [add support for source policy documents](https://github.com/cloud-gov/cg-provision/commit/42276e14f972962ef24ce3532fd9c5d5aef2b8b8)

## security considerations

The permissions are tailored to only allow access from specified accounts

### DIFF
--- a/terraform/modules/s3_bucket/kms_encrypted_bucket/bucket.tf
+++ b/terraform/modules/s3_bucket/kms_encrypted_bucket/bucket.tf
@@ -25,11 +25,6 @@ resource "aws_s3_bucket_acl" "kms_encrypted_bucket_acl" {
   acl    = var.acl
 }
 
-resource "aws_s3_bucket_policy" "kms_bucket_policy" {
-  bucket = aws_s3_bucket.kms_encrypted_bucket.id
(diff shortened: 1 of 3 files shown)


jameshochadel:
What's the motivation for the change?

markdboyd:
@jameshochadel For AWS Config, we have a bucket in one account that will receive all of the logs of config scans that happen in other accounts. And this module is being used to create that config logging bucket. So we needed a genericized way in this module to support granting other accounts access to the bucket.

We also needed to add some more policy statements to give AWS config more permissions from the source accounts to this bucket and the easiest way to do that turned out to be the [`source_policy_documents` argument](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/iam_policy_document#source_policy_documents) . It's pretty hard to explain but I can walk you through it if you want.

jameshochadel:
Awesome, that makes sense. Since it's an access control change, I wanted to understand it and have the reasoning documented for posterity. Thanks!

jameshochadel (approved):

PR #1204, opened 2023-01-06
add sq rule to allow rds to reach s3

rbogle:
- This changes gives the RDS Postgres subnet the ability to access the s3 private endpoint
- With this change customers can imoort/export data between their s3 service instance into their pg service instances. 
- This will assist with seeding or ETL of data into client databases, especially with very large datasets. 

## security considerations
None

### DIFF
--- a/terraform/modules/rds_network/sg_postgres.tf
+++ b/terraform/modules/rds_network/sg_postgres.tf
@@ -4,6 +4,12 @@
  *   vpc_id
  */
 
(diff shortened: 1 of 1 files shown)


jameshochadel (approved):
Such a great Friday project. Thanks @rbogle! 

I don't see any reason not to allow traffic between those two, and I don't think we need any SSP updates to do this.

PR #1207, opened 2023-01-11
addition of log_group for cloudwatch arn

JasonTheMain:
- addition of cloudwatch arn logs
-
-

## security considerations
None


terraform/modules/elasticsearch_broker/log_group.tf
```diff
@@ -0,0 +1,3 @@
+resource "aws_cloudwatch_log_group" "audit_log" {
+  name = "cloudwatch_arn"
```
markdboyd, line 2: This name doesn't seem very specific? Shouldn't it be something like `elasticsearch_broker_audit_log` ?

markdboyd (approved):

PR #1212, opened 2023-01-19
add security group rule allowing egress to PostgreSQL

markdboyd:

This change updates the security group to allow egress traffic to other PostgreSQL databases (port 5432) in the allowed CIDR ranges. 

[This change is intended to support "foreign data wrappers" between RDS databases](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.PostgreSQL.CommonDBATasks.Extensions.foreign-data-wrappers.html), as requested by a customer.

## security considerations

This change theoretically allows requests between databases in the same CIDR range, but username/password authentication should still be required to make the connection.


rbogle (approved):
Yay! LGTM!! 
The from port might actually be 'any' but see if this works as is.

PR #1215, opened 2023-01-20
Fix tagging collision for VPC network peering resources

markdboyd:

Due to a collision in the tags applied on the `aws_vpc_peering_connection` and `aws_vpc_peering_connection_accepter` resources, Terraform always thinks there are changes to apply for one resource, even when nothing has changed. The output from `terraform plan` looks something like:

```
  # module.stack.module.vpc_peering_parentbosh.aws_vpc_peering_connection_accepter.peer will be updated in-place
  ~ resource "aws_vpc_peering_connection_accepter" "peer" {
        id                        = <redacted>
      ~ tags                      = {
          - "deployment" = "cf-<deployment>" -> null
            # (1 unchanged element hidden)
        }
        # (8 unchanged attributes hidden)

        # (2 unchanged blocks hidden)
    }
```

This behavior is problematic because it causes our CI system to send us a message in Slack telling us to review the `terraform plan` output, when in reality nothing has changed or needs to change.

The problem turned out to be caused by the fact that the Terraform `aws_vpc_peering_connection` and `aws_vpc_peering_connection_accepter` resources actually managed a single AWS VPC peering connection resource. So defining `tags` on both Terraform resources and having default tags (`tags_all`) from different providers apply to each resource was causing Terraform to get persistently confused about which tags to apply. 

The changes in this PR update the `aws_vpc_peering_connection_accepter` Terraform resource to ignore changes to `tags` and `tags_all` (default tags), which is fine, because tags (including `Name`) are set on the `aws_vpc_peering_connection` resource. And this change resolves the persistent confusion and detection of changes by Terraform.

## security considerations

None

### DIFF
--- a/terraform/modules/vpc_peering/network.tf
+++ b/terraform/modules/vpc_peering/network.tf
@@ -29,15 +29,17 @@ resource "aws_vpc_peering_connection" "peering" {
     Name = "${var.source_vpc_id} to ${var.target_vpc_id}"
   }
 
-  lifecycle {
-    ignore_changes = [tags_all]
-  }
+
 }
 
 resource "aws_vpc_peering_connection_accepter" "peer" {
   provider                  = aws.tooling
   vpc_peering_connection_id = aws_vpc_peering_connection.peering.id
   auto_accept               = true
+
+  lifecycle {
+    ignore_changes = [tags_all]
+  }
 }
 
 # Add peering connection to target_az1_route_table with source_vpc_cidr
--- a/terraform/modules/vpc_peering/network.tf
+++ b/terraform/modules/vpc_peering/network.tf
@@ -38,6 +38,10 @@ resource "aws_vpc_peering_connection_accepter" "peer" {
   tags = {
     Name = "${var.source_vpc_id} to ${var.target_vpc_id}"
   }
+
+  lifecycle {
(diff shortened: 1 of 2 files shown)


jameshochadel (commented):
We want these resources to be tagged though, right? Do we need to introduce another method of tagging them?

markdboyd:
@jameshochadel Yeah, we want them to be tagged, but as is, this code just produces confusion because it makes the `terraform plan` always produce a diff, even though there are actually no changes to apply, which looks like this:

```
  # module.stack.module.vpc_peering_parentbosh.aws_vpc_peering_connection_accepter.peer will be updated in-place
  ~ resource "aws_vpc_peering_connection_accepter" "peer" {
        id                        = "<redacted>"
      ~ tags                      = {
          - "deployment" = "cf-staging" -> null
            # (1 unchanged element hidden)
        }
        # (8 unchanged attributes hidden)

        # (2 unchanged blocks hidden)
    }
```

I've tried to fix it and keep the `default_tags` to no avail. 

It's some permutation of this bug caused by us having multiple `provider` blocks with `default_tags`: https://github.com/hashicorp/terraform-provider-aws/issues/18311

jameshochadel (approved):
Ahh you figured it out! Fantastic news!

PR #1223, opened 2023-01-31
Expand rds subnet group

soutenniza:
- New AZ3 for rds subnet groups
- New AZ4 for rds subnet groups
- For routing reasons AZ3 will be using AZ1's route tables and AZ4 is using AZ2
- Each new subnet will have double the capacity of previous

## security considerations
None

### DIFF
--- a/terraform/modules/rds_network/network.tf
+++ b/terraform/modules/rds_network/network.tf
@@ -2,8 +2,10 @@
  * Variables required:
  *   stack_description
  *   vpc_id
- *   rds_private_cidr1
- *   rds_private_cidr2
+ *   rds_private_cidr_1
+ *   rds_private_cidr_3
+ *   rds_private_cidr_2
+ *   rds_private_cidr_4
  *   az1
  *   az2
  */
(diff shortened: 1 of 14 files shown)


markdboyd (approved):

terraform/modules/rds_network/network.tf
```diff
@@ -18,6 +20,16 @@ resource "aws_subnet" "az1_rds" {
   }
 }
 
+resource "aws_subnet" "az3_rds" {
```
jameshochadel, line 23: Can we put 3 between 2 and 4?

terraform/stacks/tooling/stack.tf
```diff
@@ -88,6 +88,8 @@ module "stack" {
   restricted_ingress_web_ipv6_cidrs      = var.restricted_ingress_web_ipv6_cidrs
   rds_private_cidr_1                     = cidrsubnet(var.vpc_cidr, 8, 20)
   rds_private_cidr_2                     = cidrsubnet(var.vpc_cidr, 8, 21)
+  rds_private_cidr_3                = cidrsubnet(var.vpc_cidr, 7, 11) # This will give 22-23
```
jameshochadel, line 91: I think this needed tflint run on it.

jameshochadel (commented):
@soutenniza Can you explain why we're doing this, for posterity? What is the motivation for the change and why are you taking this particular approach? If we have a ticket, we should link to it. (If we don't, you could also explain it there.)

PR #1240, opened 2023-03-02
improve rds sg

soutenniza:
- This only allows a singular host/ip
- Have to wrap list since it's a singluar ip so it combines correctly?

## security considerations
None


pburkholder (commented):
Do you need to update the comment at line 85?

soutenniza:
> Do you need to update the comment at line 85?

The comment still true, so good to keep it there

pburkholder (approved):

PR #1246, opened 2023-03-08
use FIPS endpoints for all providers

markdboyd:

Update all of the `provider` definitions (which there are many of in this repo) to have `use_fips_endpoint = true`, which according to the Terraform documentation should mean that FIPS endpoints are automatically used when making requests to AWS: https://registry.terraform.io/providers/hashicorp/aws/latest/docs#use_fips_endpoint. With this approach, we shouldn't need to manually specify each FIPS endpoint that we want to use but Terraform will transparently take care of using the right endpoints for us

## Security considerations

These changes update our Terraform code to use FIPS endpoints when communicating with AWS in order to comply with FIPS-140

### DIFF
--- a/terraform/modules/external_domain_broker/resources.tf
+++ b/terraform/modules/external_domain_broker/resources.tf
@@ -53,7 +53,7 @@ data "aws_canonical_user_id" "current_user" {
 }
 
 resource "aws_s3_bucket" "cloudfront_log_bucket" {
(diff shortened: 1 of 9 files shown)


jameshochadel (approved):
Excellent news that the provider has that option. Thanks Mark!

rcgottlieb (approved):

PR #1262, opened 2023-04-04
Fail CI if PR includes unformatted HCL

jameshochadel:

Terraform includes a built-in formatting tool, `terraform fmt`. This change runs the formatter on PR code and fails if it results in a dirty git index, i.e. if it results in any files being changed. Developers should configure their editors to automatically format on save using `terraform fmt`. (Editors running the [Terraform Language Server](https://github.com/hashicorp/terraform-ls) surface this as a setting.)

## security considerations
None.

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -50,7 +50,40 @@ jobs:
     params:
       path: pull-request
       status: pending
+      context: format-terraform
+  - put: pull-request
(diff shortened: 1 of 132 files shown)


jameshochadel:
Basing this on my other PR, but it'll rebase to be against `main` when that's merged.

PR #1271, opened 2023-04-27
add scope down statements for Log4JRCE WAF rule

markdboyd:

- Add dynamic scope down statements for Log4JRCE WAF rule since it was blocking legitimate customer traffic

## security considerations

None

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -528,7 +528,8 @@ jobs:
       TF_VAR_waf_regular_expressions: ((development_waf_regular_expressions))
       TF_VAR_s3_gateway_policy_accounts: ((development_s3_gateway_policy_accounts))
       TF_VAR_sns_name: cloud-gov-notifications
-      TF_VAR_known_bad_inputs_scope_down_json: ((development_known_bad_inputs_scope_down))
+      TF_VAR_scope_down_known_bad_inputs_not_match_origin_search_string: ((development_scope_down_known_bad_inputs_not_match_origin_search_string))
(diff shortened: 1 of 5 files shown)


markdboyd:
@jameshochadel Yes, I suppose it was part of the log4j mitigation, based on:

https://cloud.gov/2021/12/22/log4j_vulnerability_bod_22-02_update/
https://github.com/cloud-gov/cg-provision/pull/923/files

However, our policy to this point has been to disable any WAF rule that is blocking legitimate customer traffic. And as of today, we have a verified case of this rule blocking legitimate customer traffic. I can share details of the customer support ticket offline.

So how should we proceed?

ChrisMcGowan:
This was one of a few steps we did to help bridge the gap when this exploit started.  We also set a CF environment wide staging and running variable to disable the setting as well.  Maybe we look at stats to see how many hits we actually block with this rule, sans the legit customer issue ?  Also maybe scan all java apps on the platform again ? I think we looked for env variables the last time.

pburkholder:
The Nessus vuln scans report the full path to any log4j code on the Diego cells. I have a few scripts in https://github.com/cloud-gov/cg-scripts/tree/main/audit  (look for log4j) that can, after running `cfdot` on a Diego cell,  report back on the app, space and org. It's pretty ugly, and only works if the diego cells haven't rolled since the Nessus scan.

markdboyd:
@jameshochadel I refactored the code to remove the dynamic generation as discussed in favor of better readability/maintainability

PR #1273, opened 2023-05-02
Remove default ACL from KMS encrypted buckets

markdboyd:

AWS is now creating buckets with ACLs disabled by default: https://aws.amazon.com/blogs/aws/heads-up-amazon-s3-security-changes-are-coming-in-april-of-2023/

While this code was creating a "private" ACL for any newly created KMS-ecnrypted buckets, the effect of that is that the bucket and object owner get "full control" of the object and no one else has access rights: https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl

But a "private" ACL isn't necessary for all cases. Most of the time, we may prefer to use the new AWS default where the bucket owner only has sole access to the bucket and objects, in which case setting an ACL isn't necessary.

So this PR :

- Removes the default "private" ACL and makes an ACL optional
- Makes setting the object ownership possible in case we do want to use ACLs for a bucket

## security considerations

None


cweibel (approved):
Removes the undesired private acl and sets optional ownership

PR #1274, opened 2023-05-03
Cf timeout health check adjustments

ChrisMcGowan:
- Updating target group health checks for CF based ELBs
- Make healthy and unhealthy counts 3, increase timeout to 10 seconds, increase internal to 15 seconds

## security considerations
By increasing the timeout and interval we give secureproxy on the gorouters more time to process health checks under load and not fail causing DOS failures


cweibel (approved):
Tested in dev on a custom branch, load balancers had the desired values

PR #1278, opened 2023-05-05
add names for all security group resources

markdboyd:

- add names for all security group resources so that they have better names in AWS than `terraform-*` and their purpose is clear from their naming

## security considerations

Improving security by making the intended use of our security groups more obvious


pburkholder (approved):
This is a thing of beauty. Thank you.

PR #1279, opened 2023-05-08
Revert "add names for all security group resources (#1278)"

markdboyd:
This reverts commit 58eba25034705334f286ecc1967439143f197a51.

Unfortunately, renaming security groups is a destructive operation that will delete and recreate them. Additionally, there are long time durations to deleting security groups and lots of potential for things to go wrong:

https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group

Given the riskiness of this operation, I'm inclined to just make sure we have tags defined on our groups and not change the actual `name` property

## security considerations

None


jameshochadel (approved):
With great sadness

PR #1280, opened 2023-05-08
reverting cf elb timeout back to previous values

ChrisMcGowan:
- Reverting CF ELB health check values back to prior version to reduce number of 502s

## security considerations
Going back to previous settings lowers our 502 count and removes problem gorouters quicker from the target pool


cweibel (approved):
Put the candle back...

This reverts the degradation in behavior seen in last week's deploy for 502s

PR #1285, opened 2023-05-17
adding in IAM policy permission for masterbosh creds file upload

ChrisMcGowan:
- Master bosh creates a `creds.yml` that we need to store in S3 so policy mod to allow PUT access to the bucket for the file

## security considerations
Keeping all your secrets generated in a deploy backed-up up and secure is a good thing


cweibel (approved):
This catches any extra new variables introduced to upstream

PR #1290, opened 2023-05-24
Adding more subnets for opensearch

soutenniza:
- Changing AZ1/2 back to original subnet
- Adding AZ3/4 for new subnets

## security considerations
None


cweibel (approved):
Doing this to support expanding ip range

cweibel (approved):
Doing this to support expanding ip range

PR #1295, opened 2023-05-31
remove permissive KMS key policy

markdboyd:

- remove permissive KMS key policy

## security considerations

This improves security by removing an overly permissive statement in the KMS key policies. [This policy did come from the "default key policy" specified in the AWS documentation](https://docs.aws.amazon.com/kms/latest/developerguide/key-policy-default.html#key-policy-default-allow-root-enable-iam), however it may not be necessary for our purposes


cweibel (approved):
Removes the KMS user permissions identified in the POA&M finding even though this complies with the recommended implementation

PR #1362, opened 2023-06-05
Modification to the s3 access log bucket acl resource

dtamasang2:
- Modification to the s3 access log bucket acl resource
-
-

## security considerations
There are no security considerations for this change.


soutenniza (dismissed):

cweibel (approved):
Adds a resource dependency making sure the bucket is created first

PR #1413, opened 2023-06-14
Waf rule fix2

ChrisMcGowan:
- Updates to add TF mgmt of the WAF CloudWatch logging exclusion for a certain app that leaves the backdoor and comes back to the front causing high CloudWatch costs

## security considerations
n/a


cweibel (approved):
Adding the new WAF rules

PR #1415, opened 2023-06-14
Remove s3 audit logs trail

markdboyd:

- Remove resources to create CloudTrail for auditing S3 buckets since this is already handled by the organizational Cloudtrail

## security considerations

Since S3 bucket events are logged for all accounts by the organization CloudTrail, we don't need to set up this Cloudtrail to monitor S3 buckets per account


rcgottlieb (approved):

cweibel (approved):
Removing these since the main branch has them pulled.

PR #1416, opened 2023-06-14
Fix encryption policy

markdboyd:

- fix encryption policy typo causing S3 upload errors

## security considerations

None, the previous format was erroneous and blocking valid S3 uploads


ChrisMcGowan (approved):
Encryption looks good

PR #1417, opened 2023-06-15
Reserve fix

ChrisMcGowan:
- Add in logic to extract and output internal domain broker lb IPs to be used in bosh cloud-config in reserve range

## Notes
When you create an internal lb, AWS randomly picks IPs from each subnet you supply - you can't specify which IP to use so this code gets that IP and adds it to the reserved range in BOSH's cloud-config so BOSH does not try to use it in VM assignment.

## security considerations
n/a


soutenniza (approved):

cweibel (approved):
Add LB IP to reserved range in Cloud Config for director

PR #1420, opened 2023-06-20
Change to concourse worker role, AWS requires self reference going forward

cweibel:
- For `tooling-concourse-iaas-worker` supply an IAM role policy instead of using the default one to reference itself 
-
-

## security considerations
This is needed as a result of [https://aws.amazon.com/blogs/security/announcing-an-update-to-iam-role-trust-policy-behavior/](https://aws.amazon.com/blogs/security/announcing-an-update-to-iam-role-trust-policy-behavior/)


ChrisMcGowan (approved):
Keeping current with AWS IAM Policy changes

PR #1421, opened 2023-06-20
Force 2048 key size

cweibel:
- Should be a fix to force 2048 RSA for the newer version of certbot
-
-

## security considerations
Same cert length and type as what used to be the default


ChrisMcGowan (approved):
Force certbot to create RSA instead of ecPubkeys

PR #1529, opened 2023-07-12
rename outputs for elasticsearch broker AZs for clarity

markdboyd:

The Terraform output names and resource tags for our Elasticsearch broker subnets imply that there are 4 possible AZs - but in reality the subnets are only distributed across 2 AZs. So this PR updates the resource tags and tag output names to match reality. We can't rename the resources themselves without destroying them, which would be disruptive to customer resources.

## security considerations

None, just renaming resources and outputs


cweibel (approved):
Tags and output changes for clarity, looks good to me

cweibel (approved):
Denotes actual azs, lgtm

PR #1537, opened 2023-08-31
Removing uaa load balancer from list of LBs

cweibel:
- Remove the uaa load balancers for the list of load balancers used for cf-router-network-properties in the ops file as it now has it's own extension called cf-router-main-network-properties
-
-

## security considerations
None


ChrisMcGowan (approved):
it's all grown up and time for it to stand on it's own

PR #1547, opened 2023-09-25
Move ECR terraform

dandersonsw:
- Moves terraform for creating ECR repos from the container-scanning pipeline

## security considerations
Should not interfere with any other terraform plans, and keeps the terraform code more centrally located


jameshochadel (approved):
This is great. Do you anticipate any problems with the state file, or does this reuse the existing one?

dandersonsw:
This should reuse the existing state file, so I don't anticipate any problems

PR #1549, opened 2023-09-27
Add Autoscaler RDS

cweibel:
- Adds a module to create the RDS database needed for autoscaler in dev/stage/prod
- Allows for separate rds instance sizes for each environment to be set in cg-provision.yml
- Emits RDS outputs, similar to CF RDS, so they can be consumed by the autoscaler pipeline.

## security considerations
Passwords/hosts are passed via the state file to the pipeline for autoscaler.  The password is dynamically generated using random_string.


ChrisMcGowan (approved):
🐦 fly autoscaler fly

PR #1559, opened 2023-10-03
adding target groups for domain

soutenniza:
- Target groups for domain brokers

## security considerations
None


cweibel (approved):
Adding additional target group

PR #1561, opened 2023-10-03
hosts whole array

soutenniza:
- wrong array thing

## security considerations
none


cweibel (approved):
Flatten out the values

PR #1564, opened 2023-10-04
Add bucket & IAM roles/permissions for archiving logs opensearch

markdboyd:

- Add bucket for storing logs archive (e.g. `logs-opensearch-cf-development`)
- Add IAM profile and role to be used by archiver VM to put logs in S3 bucket

## security considerations

The IAM role/permissions are necessary for the archiver VM to write logs to this bucket. The permissions are scoped explicitly to buckets that have names starting with `logs-opensearch` so that the role will allow access to the archive buckets for each environment (e.g. `logs-opensearch-cf-development`, `logs-openearch-cf-staging`)


JasonTheMain (approved):
Looks good to me!

PR #1571, opened 2023-10-25
Iam cvd

soutenniza:
- New access key for cvd-sync
- Will keep existing until new keys are working as intended
-

## security considerations
This is reducing stale keys


ChrisMcGowan (approved):
looks good for rotating keys

PR #1572, opened 2023-10-26
Delete old cvd sync key

soutenniza:
- Confirmed new key works
- Removing old key!
-

## security considerations
One less key that is stale like 4 week old sourdough


cweibel (approved):
Removes old key

PR #1573, opened 2023-10-26
Update WAF rules

markdboyd:
Related to https://github.com/cloud-gov/private/issues/866

- Reorganized WAF rules to ensure BLOCK rules always come before rate limit rules
- Added WAF rule to block traffic by malicious JA3 fingerprints identified here: https://www.cisa.gov/news-events/cybersecurity-advisories/aa23-213a. This rule is set to **count, not block** requests
- Added WAF rule to rate limit traffic **that is not from Cloudfront by source IP** with a challenge
- Added WAF rule to rate limit traffic **that is from Cloudfront by forwarded IP** with a challenge 

These changes were already made manually on the production WAF to mitigate ongoing DDoS incidents, so this PR is just translating those changes into Terraform to make them permanent and consistent across environments

## security considerations

These changes should drastically reduce our platform's exposure to DDoS attacks and likely reduce our vulnerability to probing/sniffing attacks

### DIFF
--- a/terraform/modules/cloudfoundry/variables.tf
+++ b/terraform/modules/cloudfoundry/variables.tf
@@ -145,3 +145,9 @@ variable "waf_label_host_0" {
 variable "waf_hostname_0" {
   type = string
 }
+
+variable "nat_egress_ips" {
+  default = []
+  type    = set(string)
+  description = "IP CIDR ranges for NAT Gateways for this environment"
+}
--- a/terraform/modules/cloudfoundry/variables.tf
+++ b/terraform/modules/cloudfoundry/variables.tf
@@ -206,3 +206,8 @@ variable "api_data_gov_hosts_regex_pattern_arn" {
   type = string
   description = "ARN of regex pattern set used to identify hosts for api.data.gov"
 }
+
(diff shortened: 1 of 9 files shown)


terraform/modules/cloudfoundry/variables.tf
```diff
@@ -145,3 +145,59 @@ variable "waf_label_host_0" {
...
+variable "non_cdn_traffic_rate_limit_challenge_by_forwarded_ip" {
+  type = number
+  description = "Number of requests to allow per forwarded IP per 5 minute interval before challenging requests"
+  default = 2000
+}
+
+variable "non_cdn_traffic_rate_limit_block_by_forwarded_ip" {
+  type = number
+  description = "Number of requests to allow per forwarded IP per 5 minute interval before blocking requests"
+  default = 5000
```
markdboyd, line 192: This threshold was chosen completely arbitrarily. I figured the thresholding for blocking should be higher for challenging (2000) and I opted to be very conservative, given that our current threshold for blocking was so high it was not doing anything

JasonTheMain (approved):
:Looks good to me

PR #1575, opened 2023-11-01
Adjust waf rules

markdboyd:

- Add exception for some hosts to rate limiting by source IP
- Remove WAF rule to count requests for api.data.gov

## security considerations

Only the change the rule to add exceptions for rate limiting by source IP has meaningful security impact. But at this time, we have to make exceptions for certain customers whose traffic legitimately exceeds the rate limit which is effective for all other traffic


cweibel (approved):
Looks good

PR #1584, opened 2023-11-09
Iam stale - last changes for gov/com

ChrisMcGowan:
- commenting out unused federalist-auditor account

## Notes

I'll circle back in a few weeks pull all the commented out code for final removal

## security considerations
Removing stale accounts helps tighten security

### DIFF
--- a/terraform/stacks/tooling/outputs.tf
+++ b/terraform/stacks/tooling/outputs.tf
@@ -616,28 +616,29 @@ output "staging_doomsday_lb_target_group" {
 #  sensitive = true
 #}
 
-/* federalist auditor user */
-output "federalist_auditor_username" {
-  value = module.federalist_auditor_user.username
-}
-
-output "federalist_auditor_access_key_id_prev" {
-  value = module.federalist_auditor_user.access_key_id_prev
-}
-
-output "federalist_auditor_secret_access_key_prev" {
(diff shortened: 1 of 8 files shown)


terraform/stacks/tooling/outputs.tf
```diff
@@ -616,28 +616,29 @@ output "staging_doomsday_lb_target_group" {
...
-
-output "federalist_auditor_access_key_id_curr" {
-  value = module.federalist_auditor_user.access_key_id_curr
-}
-
-output "federalist_auditor_secret_access_key_curr" {
-  value     = module.federalist_auditor_user.secret_access_key_curr
-  sensitive = true
-}
+# Not in use - McGowan 11/9/2023
```
markdboyd, line 619: Can we not just delete vs commenting? We always have Git history to allow us to restore things if necessary

ChrisMcGowan: The commenting was done on purpose for quick reversal especially on accounts in higher environments that don't exist in dev for testing.  Yes it's an extra PR to remove then as another step but it allows easy reversal without looking through git history.

cweibel (approved):
Dealers choice on comment v commit for easier rollback for untestable impact removing this will have

PR #1586, opened 2023-11-27
Removing old comments

ChrisMcGowan:
- NoOp change - just removing old commented out code and a module folder no longer in use

## security considerations
Keeping code base clean and current and removing old iam-stale key items


markdboyd (approved):
I love the smell of deleted code in the morning

PR #1613, opened 2024-03-15
Switch to list of ip cidrs to block

cweibel:
- Switch for a single cidr range to a list of cidr ranges which will dynamically be created starting at rule 20
- This solution has a maximum of 80 cidr ranges without intervention
- Tested manually against dev

## security considerations
None, end result is the exact same acl rules just future proofing if we need to expand the list of ip cidrs

### DIFF
--- a/terraform/modules/bosh_vpc/variables.tf
+++ b/terraform/modules/bosh_vpc/variables.tf
@@ -67,6 +67,8 @@ variable "s3_gateway_policy_accounts" {
 }
 
 #Placeholder for real value, passed as a secret
-variable "block_range_20" {
-  default = "192.168.0.0/32"
+variable "cidr_blocks" {
+  type    = list(string)
+  default = ["192.168.0.0/32", "192.168.0.1/32"]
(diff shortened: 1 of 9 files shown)


terraform/modules/bosh_vpc/variables.tf
```diff
@@ -67,6 +67,8 @@ variable "s3_gateway_policy_accounts" {
 }
 
 #Placeholder for real value, passed as a secret
-variable "block_range_20" {
-  default = "192.168.0.0/32"
-}
\ No newline at end of file
+variable "cidr_blocks" {
+  type    = list(string)
+  default = ["192.168.0.0/32", "192.168.0.1/32"]
```
markdboyd, line 72: I don't think we want to set any default values. Because if we happened to not pass the real values through in CI automatically, then traffic from these default CIDRs would get blocked

markdboyd (dismissed):

ChrisMcGowan (approved):

PR #1642, opened 2024-04-18
Make fingerprint rules dynamic

markdboyd:

- Make fingerprint rules truly dynamic so we don't need to use dummy values in our variable to deploy and so that the rules will be totally conditional on how many values there are

## security considerations

Doesn't change the behavior of the block by fingerprint rules, it just makes them totally conditional so we can easily remove or update them just by changing the input variables


JasonTheMain (dismissed):
Looks good to me

soutenniza (approved):

PR #1644, opened 2024-04-22
Add an IAM user for the Cloud Service Broker and the SMTP brokerpak

jameshochadel:

- Code based on previous work by data.gov: https://github.com/GSA-TTS/datagov-brokerpak-smtp/blob/main/permission-policies.tf
- Related to https://github.com/cloud-gov/product/issues/2988
- Related to https://github.com/cloud-gov/product/issues/2989

## security considerations

The new user has access that might be better scoped down, but not having run the brokerpak yet, I'm not sure exactly what it needs. I will make a note to use the IAM Access Analyzer to scope down the permissions after testing — especially the two FullAccess policies. Issue here: https://github.com/cloud-gov/product/issues/2993

### DIFF
--- a/terraform/modules/csb/iam.tf
+++ b/terraform/modules/csb/iam.tf
@@ -1,0 +1,86 @@
+// Originally from https://github.com/GSA-TTS/datagov-brokerpak-smtp/blob/main/permission-policies.tf
+locals {
+  this_aws_account_id = data.aws_caller_identity.current.account_id
+}
+
+data "aws_caller_identity" "current" {}
+
+data "aws_iam_policy_document" "brokerpak_smtp" {
+  statement {
+    effect    = "Allow"
+    actions   = ["ses:*"]
+    resources = ["*"]
+  }
+
+  statement {
+    effect = "Allow"
+    actions = [
+      "iam:CreateUser",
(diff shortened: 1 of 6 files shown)


terraform/modules/csb/iam.tf
```diff
@@ -0,0 +1,86 @@
...
+
+    // Route53 manager: for aws_route53_record, aws_route53_zone
+    "arn:aws-us-gov:iam::aws:policy/AmazonRoute53FullAccess",
+
+    // SMTP brokerpak policy defined above
+    "arn:aws-us-gov:iam::${local.this_aws_account_id}:policy/${aws_iam_policy.brokerpak_smtp.name}",
+  ])
+
+  user       = aws_iam_user.iam_user.name
+  policy_arn = each.key
```
markdboyd, line 85: I'm surprised this wouldn't be `each.value`, because I thought `each.key` would be the index of the value in the array

jameshochadel: Since I'm using `toset()`, a set is passed, and its values are referenced by `each.key`: https://developer.hashicorp.com/terraform/language/meta-arguments/for_each#basic-syntax

markdboyd (approved):

PR #1651, opened 2024-04-24
Update resource arn for elasticsearch logs

hursey013:
- Updates resource policy to use `arn:aws-us-gov:logs` as opposed to `arn:aws:logs`

## security considerations
None

### DIFF
--- a/terraform/modules/cloudwatch/policies.tf
+++ b/terraform/modules/cloudwatch/policies.tf
@@ -7,7 +7,7 @@ data "aws_iam_policy_document" "elasticsearch-log-publishing-policy" {
       "logs:PutLogEvents",
     ]
 
-    resources = ["arn:aws:logs:*"]
+    resources = ["arn:${var.aws_partition}:logs:*"]
 
     principals {
       identifiers = ["es.amazonaws.com"]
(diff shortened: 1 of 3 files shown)


jameshochadel (dismissed):
Nice catch!

terraform/modules/cloudwatch/policies.tf
```diff
@@ -7,7 +7,7 @@ data "aws_iam_policy_document" "elasticsearch-log-publishing-policy" {
       "logs:PutLogEvents",
     ]
 
-    resources = ["arn:aws:logs:*"]
+    resources = ["arn:aws-us-gov:logs:*"]
```
markdboyd, line 10:
```suggestion
    resources = ["arn:${var.aws_partition}:logs:*"]
```

I would add `aws_partition` as a variable to this module. Then update this block in `stacks/main/stack.tf` like so:

```
module "cloudwatch" {
  source = "../../modules/cloudwatch"

  stack_description                   = var.stack_description
  cg_platform_notifications_arn       = module.sns.cg_platform_notifications_arn
  cg_platform_slack_notifications_arn = module.sns.cg_platform_slack_notifications_arn
  load_balancer_dns                   = module.cf.lb_arn_suffix
  aws_partition =  data.aws_partition.current.partition
}
```

It keeps us from hardcoding a value and will ensure it is deployed with the correct value

hursey013: Good call!

markdboyd (approved):

PR #1657, opened 2024-05-09
pages: Add redis and postgres images to ECR

apburnes:
- Adds redis v7.2 
- Adds postgres v15 

## security considerations
Adds redis and postgres registries for hardened images


jameshochadel (approved):
We should chat about the versioning thing Pages is doing at some point, but for now this looks fine

PR #1658, opened 2024-05-09
sort cidr blocks for security group to avoid unnecessary TF changes

markdboyd:

Terraform sometimes shows that there are pending changes to the CIDR blocks for a security group when in reality it is just reordering them. I think that by sorting the CIDR blocks input to the security group, it should cause Terraform to always apply them in a consistent order and avoid confusing `plan` output that implies that things are changing when in fact they're just being reorganized.

## security considerations

None, just addressing some confusing Terraform behavior


jameshochadel (approved):
Sold!

PR #1670, opened 2024-05-15
Refactor s3 lifecycle expiration blocks

markdboyd:

- Remove `filter { prefix = "" }` blocks since the default value of `prefix` is `""`
- For S3 lifecycle configurations where expiration is the only option, use `count` to not create the resource if no expiration is provided
- Use dynamic TF blocks to generate `expiration` blocks conditionally instead of relying on buggy TF behavior (https://github.com/terraform-providers/terraform-provider-aws/issues/1402)

## security considerations

None, just refactoring TF for clarity and reliability


soutenniza (approved):

JasonTheMain (dismissed):
Looks good

PR #1692, opened 2024-07-11
Fix attribute not existing until apply

jameshochadel:

- See inline comment

## security considerations
None

### DIFF
--- a/terraform/modules/csb/iam.tf
+++ b/terraform/modules/csb/iam.tf
@@ -75,15 +75,22 @@ resource "aws_iam_access_key" "iam_access_key" {
   user = aws_iam_user.iam_user.name
 }
 
+locals {
+  // Attribute aws_iam_policy.brokerpak_smtp.arn is not determined until apply, so it cannot be
+  // referenced in for_each below. Build the ARN here instead.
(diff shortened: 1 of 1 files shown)


terraform/modules/csb/iam.tf
```diff
@@ -75,15 +75,22 @@ resource "aws_iam_access_key" "iam_access_key" {
   user = aws_iam_user.iam_user.name
 }
 
+locals {
+  // Attribute aws_iam_policy.brokerpak_smtp.arn is not determined until apply, so it cannot be
+  // referenced in for_each below. Build the ARN here instead.
+  brokerpak_smtp_arn = "arn:aws-us-gov:iam::${local.this_aws_account_id}:policy/${aws_iam_policy.brokerpak_smtp.name}"
```
markdboyd, line 81:
```suggestion
  brokerpak_smtp_arn = "arn:aws-us-gov:iam::${local.this_aws_account_id}:policy/${aws_iam_policy.brokerpak_smtp.name}"
```

I assume that `aws-us-gov` should be a variable so this could be deployed in gov or commercial? Since it's used possibly for either below

jameshochadel: Whoops, good catch. Updating...

markdboyd (dismissed):

markdboyd (approved):

PR #1700, opened 2024-07-22
Rds engine change

rcgottlieb:
- Modified tf to upgrade db engine to Postgres 15.5
- Update parameter group to the correct family of Postgres

## security considerations
None


cweibel (approved):
Paired with Robert on these changes

PR #1734, opened 2024-08-15
pulling out the external-domain-broker key rotation code as it won't …

rcgottlieb:
- Pulling out all of the code associated with external domain broker key rotation
- There are many issues with the particular block I chose to use for this
- The goal is to manually rotate a few of these and see if there is a pattern that will inform automation
 
## security considerations
None


JasonTheMain (approved):
Looksgood

PR #1739, opened 2024-08-20
Fix external domain broker code

markdboyd:

Related to https://github.com/cloud-gov/private/issues/1098

Fixes for #1735. Those changes are failing with:

```
LimitExceeded: Maximum policy size of 2048 bytes exceeded
```

So this PR splits one IAM policy into multiple policies to avoid hitting that limit. It also adds an output for the rule group ARN that is needed for reference elsewhere.

## security considerations

None, just refactoring the IAM policies to address errors.


JasonTheMain (approved):
Looks good!

PR #1745, opened 2024-08-27
update perms for external domain broker 

markdboyd:

Related to https://github.com/cloud-gov/private/issues/1098

- Add permission to tag IAM certificates in GovCloud
- Fix permissions for Route53 health check creation so that creation can succeed
- Add restriction on all IAM actions in `external_domain_broker` module to only allow incoming requests from the NAT egress IPs of the corresponding GovCloud stack
- Refactor `external_domain_broker_govcloud` TF code so that it can be effectively scanned by TF linting tools like [trivy](https://github.com/aquasecurity/trivy)
- Refactor `external_domain_broker_govcloud` permissions to be more explicit and least privilege

## security considerations

These changes enforce tighter restrictions on the IAM policies used by the external domain broker, which is good for security. Specifically we now validate the source IP of requests to the external domain broker IAM permissions in the commercial partition.


ci/pipeline.yml
```diff
@@ -1585,10 +1594,6 @@ jobs:
             resource: terraform-yaml-external-production
       - task: check-certificates
         image: general-task
-        file: cg-provision-repo/ci/check-certificates.yml
```
markdboyd, line 1588: This change is unrelated to the rest but the pipeline file was failing local YAML validation for me and this block seemed to be the cause. Seems to have come from #1696. Paging @bengerman13

JasonTheMain (dismissed):
Looks good Mark

terraform/modules/external_domain_broker_govcloud/resources.tf
markdboyd, line 1: Renamed to `terraform/modules/external_domain_broker_govcloud/iam.tf`

PR #1748, opened 2024-08-29
add external-domain-broker image

dandersonsw:
- Adds an ECR repo for the image used by the external-domain-broker repo

## security considerations
Adding ECR repo for external-domain-broker

### DIFF
--- a/terraform/stacks/ecr/stack.tf
+++ b/terraform/stacks/ecr/stack.tf
@@ -17,7 +17,7 @@ variable "repositories" {
     "cron-resource",
     "csb",
     "email-resource",
-    "external-domain-broker",
+    "external-domain-broker-testing",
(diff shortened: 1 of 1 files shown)


terraform/stacks/ecr/stack.tf
```diff
@@ -17,6 +17,7 @@ variable "repositories" {
     "cron-resource",
     "csb",
     "email-resource",
+    "external-domain-broker",
```
markdboyd, line 20:
```suggestion
    "external-domain-broker-testing",
```

markdboyd: This image is only used for testing in our pipeline, so it might be helpful if the name clarifies that

markdboyd (approved):

PR #1749, opened 2024-09-03
Adding 3 additional outputs for s3 ip cidr ranges

cweibel:
- Adds 3 additional ip cidr ranges for S3.  This is then used as an input to cg-deploy-cf/terraform/stack/asg.tf
- Related to https://github.com/cloud-gov/cg-deploy-cf/pull/881
-

## security considerations
Makes publicly published ip ranges available as terraform output variables.

### DIFF
--- a/terraform/stacks/main/outputs.tf
+++ b/terraform/stacks/main/outputs.tf
@@ -404,6 +404,18 @@ output "s3_gateway_endpoint_cidr_2" {
   value = data.aws_prefix_list.s3_gw_cidrs.cidr_blocks[1]
 }
 
+output "s3_gateway_endpoint_cidr_3" {
+  value = data.aws_prefix_list.s3_gw_cidrs.cidr_blocks[2]
+}
+
+output "s3_gateway_endpoint_cidr_4" {
+  value = data.aws_prefix_list.s3_gw_cidrs.cidr_blocks[3]
+}
+
+output "s3_gateway_endpoint_cidr_5" {
+  value = data.aws_prefix_list.s3_gw_cidrs.cidr_blocks[4]
+}
+
 /* s3 private gateway endpoint ips for public egress */
 output "vpc_endpoint_customer_s3_if1_ip" {
   value = module.stack.vpc_endpoint_customer_s3_if1_ip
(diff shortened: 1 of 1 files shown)


terraform/stacks/main/outputs.tf
```diff
@@ -404,6 +404,18 @@ output "s3_gateway_endpoint_cidr_2" {
...
+}
+
+output "s3_gateway_endpoint_cidr_4" {
+  value = data.aws_prefix_list.s3_gw_cidrs.cidr_blocks[3]
+}
+
+output "s3_gateway_endpoint_cidr_5" {
+  value = data.aws_prefix_list.s3_gw_cidrs.cidr_blocks[4]
+}
+
```
markdboyd, line 418: We should replace all of the `s3_gateway_endpoint_cidr_*` blocks with:

```
output "s3_gateway_endpoint_cidrs" {
    value = data.aws_prefix_list.s3_gw_cidrs.cidr_blocks[*]
}
```

Otherwise, we'll still hardcoding an expectation of how many IPs there will be for the gateway endpoint and could run into the same problem again

terraform/stacks/main/outputs.tf
```diff
@@ -396,12 +396,8 @@ output "elasticsearch_security_group" {
...
 
 /* S3 Gateway Endpoint CIDRs for CF ASGs*/
-output "s3_gateway_endpoint_cidr_1" {
-  value = data.aws_prefix_list.s3_gw_cidrs.cidr_blocks[0]
-}
-
-output "s3_gateway_endpoint_cidr_2" {
-  value = data.aws_prefix_list.s3_gw_cidrs.cidr_blocks[1]
+output "s3_gateway_endpoint_cidrs" {
+  value = data.aws_prefix_list.s3_gw_cidrs.cidr_blocks[*]
```
markdboyd:
```suggestion
  value = data.aws_prefix_list.s3_gw_cidrs.cidr_blocks
```

Upon further review, I don't even think we need the `[*]`. The `cidr_blocks` attribute is already a list of IP CIDRs: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/prefix_list#attribute-reference

markdboyd (approved):

PR #1769, opened 2024-09-16
Handle when module count is 0

jameshochadel:
- The module is currently only created in dev; this handles other environments.

## security considerations
None.

### DIFF
--- a/terraform/stacks/main/outputs.tf
+++ b/terraform/stacks/main/outputs.tf
@@ -751,19 +751,19 @@ output "csb" {
   sensitive   = true
   value = {
     ecr_user = {
-      username               = module.csb_broker[0].ecr_user_username
-      access_key_id_curr     = module.csb_broker[0].ecr_user_access_key_id_curr
-      secret_access_key_curr = module.csb_broker[0].ecr_user_secret_access_key_curr
-      access_key_id_prev     = module.csb_broker[0].ecr_user_access_key_id_prev
-      secret_access_key_prev = module.csb_broker[0].ecr_user_secret_access_key_prev
+      username               = length(module.csb_broker[0]) > 0 ? module.csb_broker[0].ecr_user_username : ""
+      access_key_id_curr     = length(module.csb_broker[0]) > 0 ? module.csb_broker[0].ecr_user_access_key_id_curr : ""
+      secret_access_key_curr = length(module.csb_broker[0]) > 0 ? module.csb_broker[0].ecr_user_secret_access_key_curr : ""
+      access_key_id_prev     = length(module.csb_broker[0]) > 0 ? module.csb_broker[0].ecr_user_access_key_id_prev : ""
+      secret_access_key_prev = length(module.csb_broker[0]) > 0 ? module.csb_broker[0].ecr_user_secret_access_key_prev : ""
     }
     rds = {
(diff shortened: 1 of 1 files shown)


terraform/stacks/main/outputs.tf
```diff
@@ -751,19 +751,19 @@ output "csb" {
   sensitive   = true
   value = {
     ecr_user = {
-      username               = module.csb_broker[0].ecr_user_username
-      access_key_id_curr     = module.csb_broker[0].ecr_user_access_key_id_curr
-      secret_access_key_curr = module.csb_broker[0].ecr_user_secret_access_key_curr
-      access_key_id_prev     = module.csb_broker[0].ecr_user_access_key_id_prev
-      secret_access_key_prev = module.csb_broker[0].ecr_user_secret_access_key_prev
+      username               = length(module.csb_broker[0]) > 0 ? module.csb_broker[0].ecr_user_username : ""
```
markdboyd, line 754:
```suggestion
      username               = length(module.csb_broker) > 0 ? one(module.csb_broker).ecr_user_username : ""
```

`module.csb_broker[0]` would be an object itself, not an array, I think? The `one` thing I think is just nice, but less important

jameshochadel: Good catch, thanks. It looks like I can simplify this even further to just `one(module.csb_broker[*].property)...` since `one` returns null if the collection contains no elements.

terraform/stacks/main/outputs.tf
```diff
@@ -751,19 +751,19 @@ output "csb" {
   sensitive   = true
   value = {
     ecr_user = {
-      username               = module.csb_broker[0].ecr_user_username
-      access_key_id_curr     = module.csb_broker[0].ecr_user_access_key_id_curr
-      secret_access_key_curr = module.csb_broker[0].ecr_user_secret_access_key_curr
-      access_key_id_prev     = module.csb_broker[0].ecr_user_access_key_id_prev
-      secret_access_key_prev = module.csb_broker[0].ecr_user_secret_access_key_prev
+      username               = one(module.csb_broker[*].ecr_user_username)
```
markdboyd, line 754:
```suggestion
      username               = one(module.csb_broker[*]).ecr_user_username
```

I think the right parenthesis is in the wrong place?

jameshochadel: Shoot, yes, sorry. Sloppy work

markdboyd: Also, I don't think will work as is. From what I read, `one` will

> If the collection is empty, one returns null. Otherwise, one returns the first element. If there are two or more elements then one will return an error.

And `.ecr_user_username` on `null` would return an error.

jameshochadel: I was also suspicious of this, but I wrote it the same way they wrote it in the docs; if the terraform validates, perhaps I try it? At the risk of another PR, I suppose.

markdboyd: The TF community seems to agree?

https://discuss.hashicorp.com/t/terraform-outputs-with-count-index/32555/2

I guess TF converts `module.csb_broker[*].ecr_user_username` to an empty array of an array of values

And `one` returns the first value or `null`

markdboyd (approved):

PR #1783, opened 2024-09-25
Update dependabot

markdboyd:

- Update dependabot config to scan all directories of the current code structure

## security considerations

Allows Dependabot to ensure that we're using the latest/patched versions of Terraform providers throughout our code


scripts/update-dependabot.sh
JasonTheMain, line 1: Should this be empty or is it just appearing empty?

JasonTheMain: saw that github was just being weird

JasonTheMain (approved):
Looks good

PR #1786, opened 2024-09-26
Fix ECR tag detection

markdboyd:

- Add provider with default tags so that it does not always detect changes on tags
- True up database version for Defect Dojo

## security considerations

None, just improving Terraform


dandersonsw (approved):
Looks good, ty!

soutenniza (approved):

PR #1815, opened 2024-12-26
Permit ECR pull users to pull from multiple repositories

jameshochadel:

And grant the CSB user access to its companion repository, csb-docproxy.

Also switch from the templatefile data source to the templatefile function, at the recommendation of the terraform docs. https://registry.terraform.io/providers/hashicorp/template/latest/docs/data-sources/file

## security considerations

Expands the permissions of the user that Cloud Foundry uses to pull the CSB-related images.


terraform/modules/iam_user/ecr_pull_user/policy.tftpl
cweibel, line 1: That's nifty

cweibel (approved):
lgtm

PR #1821, opened 2025-01-07
Add account-wide alarms for SES sender reputation

jameshochadel:

- Alarms trigger at warning and critical levels if SES reputation metrics are in danger of putting our account under review
- The platform operations team is alerted through our usual notification channels
- Related to https://github.com/cloud-gov/product/issues/3214

## security considerations

Follows standard approach for CloudWatch alarms.


terraform/modules/cloudwatch/alarms.tf
```diff
@@ -71,3 +71,137 @@ resource "aws_cloudwatch_metric_alarm" "load_balancer_request_spike" {
...
+  alarm_name        = "SES-BounceRate-Warning-AccountWide"
+  alarm_description = "Warning: The bounce rate for this AWS account has exceeded 2%. The Platform team should take action so AWS does not pause our ability to send email. Runbook link: https://github.com/cloud-gov/internal-docs/blob/main/docs/runbooks/Brokered-Services/AWS-SES.md"
+
+  metric_query {
+    id = "m1"
+    metric {
+      metric_name = "BounceRate"
+      namespace   = "AWS/SES"
+      period      = 300
+      stat        = "Average"
```
bengerman13, line 92: question for all of these, I think:
is the `BounceRate` cumulative or bucketed? When AWS evaluates whether we're good senders, what window are they looking at?

(reading AWS's docs it sounds like the answer to the second question is "maybe")

jameshochadel: The best answer I can find is from this page, which says:

> We calculate your bounce rate based on a representative volume of email. A representative volume is an amount of email that represents your typical sending practices. To be fair to both high- and low-volume senders, the representative volume is different for each account and changes as the account's sending patterns change.

So it sounds like they sample delivery notifications. That doesn't explain if the thresholds are instantaneous or averaged over some period of time, though. Are there changes you'd like to see here if it's one or the other?

bengerman13: if it's an average, we'd want to watch trends very closely. I think we have to assume this is windowed, and hope it's roughly the same window/metric they actually evaluate for blocks

bengerman13 (approved):

PR #1823, opened 2025-01-22
Ignore all WAF changes due to false positives

jameshochadel:

Due to a bug in the provider, terraform detects changes in every rule during every plan and apply. The apply takes ~10 minutes, unnecessarily slowing iteration on the rest of the codebase.

This change works around the issue by ignoring all changes to WAF rules, legitimate or not. To make changes to WAF rules in the future, developers will have to remove 'rule' from this block.

- Related to: https://github.com/hashicorp/terraform-provider-aws/issues/33124
- Slack discussion: https://gsa-tts.slack.com/archives/C0ENP71UG/p1737582745012769

## security considerations
None.


cweibel (approved):
LGTM, chatted about this in #cg-platform as an experiment to determine behavior

bengerman13 (approved):

PR #1826, opened 2025-02-10
create new key for rotation for iam cert provision

rcgottlieb:
- First step in rotating the iam cert provision keys
-
-

## security considerations
None


cweibel (approved):
LGTM, would be curious what downstream is looking for these output values.

Assuming there would be a second PR which removes the v3 ref once cycled out from anything currently using it to the v3 values

PR #1830, opened 2025-02-12
Moving pipeline to use GH token from credhub

ChrisMcGowan:
- Move GH token for PR check out of s3 file and into credhub for centralized management

## security considerations
Mange GH token lifecycle in credhub prevents duplicate data and split-brain during token rotation


cweibel (approved):
-/_

PR #1832, opened 2025-02-18
Simplify IAM resources for CSB project

jameshochadel:

The original iteration of the module used some clever Terraform tricks to create two IAM users for the broker from the same set of TF resources, one in GovCloud and Commercial, with different IAM policies. As the CSB project has grown, it's required more users: One to pull ECR images, one to deploy the broker itself via Concourse, and soon, one for the CSB Helper so it can confirm its SNS subscription. The original method doesn't make sense for these users, most of which exist only in GovCloud. It has become clumsy to maintain.

This commit cleans up the internal terraform and the stack outputs related to the CSB. It also normalizes the IAM user naming and drops a few unused outputs.

## security considerations

None. Just a refactoring.


JasonTheMain (approved):
Looks good to me

PR #1885, opened 2025-03-24
ccdb force_ssl

cweibel:
- Adding support to configure `rds.force_ssl` from the pipeline for each environment's CCDB
- This should be a no-op change, default to 1, override to 0 in the pipeline (which reflects what the parameter group currently has)
- Part of https://github.com/cloud-gov/private/issues/2392
-

## security considerations
This is a no-op change.  A subsequent PR will enable enforcement per environment.


jameshochadel (approved):
Looks good!

PR #1907, opened 2025-03-31
Adding generic update db script

cweibel:
- Creating a generic `update db` script which will take in the list of dbs, host, user and pass from the pipeline instead of an almost identical set of `update-blah-db.sh` scripts.  Less code to maintain.  Will follow the same pattern as `STATE_FILE_PATH`.
- There will be a separate PR for `pipeline.yml` as it is verified this works.
- Part of https://github.com/cloud-gov/private/issues/2406
-

## security considerations
No changes to security, moving around where variables are used


ChrisMcGowan (approved):
:shipit:

PR #1908, opened 2025-03-31
Clear up variable names

cweibel:
- Modified variable names so it is clearer that what is passed in via pipeline.yml params is the name of the terraform variable, not the value of the terraform variable.
- Part of https://github.com/cloud-gov/private/issues/2406
-

## security considerations
None, script is not yet used, however still is just moving around where variables are set.


ChrisMcGowan (approved):
:shipit:

PR #1924, opened 2025-04-25
Opensearch s3 bucket

JasonTheMain:
- add in iam for opensearch s3 role
-
-

## security considerations
None


terraform/modules/iam_user/s3_opensearch/outputs.tf
```diff
@@ -0,0 +1,19 @@
...
+  value = ""
+}
+
+output "access_key_id_curr" {
+  value = aws_iam_access_key.iam_access_key_v1.id
+}
+
+output "secret_access_key_curr" {
+  value = aws_iam_access_key.iam_access_key_v1.secret
+}
```
markdboyd, line 19: These need to be outputs of the main stack so we can read them, like so:

https://github.com/cloud-gov/terraform-provision/blob/ef28b7234d43d70bd086c48d4b921632b5dab268/terraform/stacks/main/outputs.tf#L690-L697

I'd call the output `logs_opensearch_s3_user_access_key_id_curr`, ec

terraform/modules/cloudfoundry/buckets.tf
```diff
@@ -39,3 +39,12 @@ module "logs-opensearch-archive" {
   acl              = "private"
   object_ownership = "ObjectWriter"
 }
+
+module "logs-opensearch-cf-audit-events" {
+  source           = "../s3_bucket/log_encrypted_bucket"
+  bucket           = "logs-opensearch-cf-audit-events-${var.stack_prefix}"
+  aws_partition    = var.aws_partition
+  expiration_days  = 372 # 31 days * 12 months = 372 days
```
markdboyd, line 47: We need to keep things in S3 for at least 30 months, so ~930 days

JasonTheMain: interesting, i thought because opensearch?

markdboyd (approved):

PR #1925, opened 2025-04-29
Removing logsearch static ips :)

soutenniza:
- logsearch is no longer using static ips ^^
-
-

## security considerations
None, this makes deployments more flexible


ChrisMcGowan (approved):
:shipit:

PR #1931, opened 2025-05-12
Separate out STIG RDS changes to apply only to csb/stig_broker

pburkholder:

In discussion w/ @jameshochadel , this split out STIG hardening into the following copies of 
the original un-stig'd code:

- terraform/modules/csb/broker_stig
- terraform/modules/rds_stig

and updates [terraform/stacks/main/stack.tf] to add:

-  resource "random_password" "csb_rds_stig_password"
- module "csb_stig_broker" 

## security considerations
Safe. Adds no new functionality or secrets

### DIFF
--- a/terraform/modules/csb/broker_stig/versions.tf
+++ b/terraform/modules/csb/broker_stig/versions.tf
@@ -1,0 +1,13 @@
+terraform {
+  required_version = ">= 0.15"
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "< 6.0.0"
+    }
+    cloudfoundry = {
+      source  = "cloudfoundry-community/cloudfoundry"
+      version = "< 1.0"
+    }
+  }
+}
--- a/terraform/modules/rds_stig/database.tf
+++ b/terraform/modules/rds_stig/database.tf
@@ -1,0 +1,44 @@
+resource "aws_db_instance" "rds_database" {
+  engine         = var.rds_db_engine
+  engine_version = var.rds_db_engine_version
+
+  multi_az = var.rds_multi_az
+
+  lifecycle {
+    ignore_changes  = [identifier]
+    prevent_destroy = true
+  }
+
+  identifier = "${var.stack_description}-${element(split("-", uuid()), 4)}"
+
+  final_snapshot_identifier = var.rds_final_snapshot_identifier == "" ? "final-snapshot-${var.rds_db_name}-${var.stack_description}" : var.rds_final_snapshot_identifier
+
+  backup_retention_period = 35
+
+  auto_minor_version_upgrade = true
+
+  db_name           = var.rds_db_name
+  allocated_storage = var.rds_db_size
+  storage_type      = var.rds_db_storage_type
+  iops              = var.rds_db_iops
+  instance_class    = var.rds_instance_type
+
+  username = var.rds_username
+  password = var.rds_password
+
+  storage_encrypted = true
+
+  db_subnet_group_name   = var.rds_subnet_group
+  vpc_security_group_ids = var.rds_security_groups
+
+  performance_insights_enabled = var.performance_insights_enabled
+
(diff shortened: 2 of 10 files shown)


terraform/stacks/main/stack.tf
```diff
@@ -467,6 +467,7 @@ resource "random_password" "csb_rds_password" {
 }
 
 module "csb_broker" {
+  count             = var.stack_description != "development" ? 2 : 0
```
pburkholder, line 470: @jameshochadel Is this right? Or does that create 2 copies of the broker in each non-development env?

markdboyd: This would create 2 copies of it. If you're trying to create this resource conditionally, you would want to use `1`, which means: if the condition is true, then create 1 instance of this resource, otherwise 0.

terraform/modules/csb/broker_stig/versions.tf
```diff
@@ -0,0 +1,13 @@
...
+  required_version = ">= 0.15"
+  required_providers {
+    aws = {
+      source  = "hashicorp/aws"
+      version = "< 6.0.0"
+    }
+    cloudfoundry = {
+      source  = "cloudfoundry-community/cloudfoundry"
+      version = "< 1.0"
+    }
```
markdboyd, line 11: I don't think we're using this provider anywhere in the code?

terraform/modules/rds_stig/database.tf
```diff
@@ -0,0 +1,44 @@
...
+  password = var.rds_password
+
+  storage_encrypted = true
+
+  db_subnet_group_name   = var.rds_subnet_group
+  vpc_security_group_ids = var.rds_security_groups
+
+  performance_insights_enabled = var.performance_insights_enabled
+
+  parameter_group_name = var.rds_db_engine == "postgres" ? join("", aws_db_parameter_group.parameter_group_postgres.*.id) : join("", aws_db_parameter_group.parameter_group_mysql.*.id)
```
markdboyd, line 36:
```suggestion
  parameter_group_name = var.rds_db_engine == "postgres" ?  aws_db_parameter_group.parameter_group_postgres.[0].id :  aws_db_parameter_group.parameter_group_mysql.[0].id
```

If there is only one of each type of resource, which I believe there is, then I think it's more obvious for the code to reference that one instance

jameshochadel: I think you need to keep the original `*` operator, because the proposed change will fail if there are zero instances created by the `count`.

terraform/modules/csb/broker_stig/versions.tf
jameshochadel: And if you were, you'd want to use `cloudfoundry/cloudfoundry`, not the older community version

terraform/modules/rds_stig/database.tf
pburkholder: Sure, but should but my formatter complains, so I think it should be: 

```
aws_db_parameter_group.parameter_group_postgres[0].id :
```
Not:

```
aws_db_parameter_group.parameter_group_postgres.[0].id :
```

Right?

markdboyd (dismissed):

terraform/stacks/main/outputs.tf
```diff
@@ -783,12 +783,12 @@ output "csb" {
       }
     }
     rds = {
-      host     = module.csb_broker.rds_host
-      port     = module.csb_broker.rds_port
-      url      = module.csb_broker.rds_url
-      name     = module.csb_broker.rds_name
-      username = module.csb_broker.rds_username
-      password = module.csb_broker.rds_password
+      host     = module.csb_broker[0].rds_host
```
pburkholder, line 786: How do I modify outputs.tf so it uses the `csb_broker_stig` code in the dev environment (instead of `csb_broker`)

pburkholder:
Moved work over to #1962 to avoid mucking up the existing CSB DBs (until we're ready)

PR #1932, opened 2025-05-12
Add MySQL hardening SQL script

pburkholder:
-
- Models mysql updates after psql
-

## security considerations
Safe. Provides hardening..

### DIFF
--- a/ci/scripts/create-and-update-mysql.sh
+++ b/ci/scripts/create-and-update-mysql.sh
@@ -9,6 +9,7 @@ db_address=$(${TERRAFORM} output -raw -state="${STATE_FILE_PATH}" "${TERRAFORM_D
 db_user=$(${TERRAFORM} output -raw -state="${STATE_FILE_PATH}" "${TERRAFORM_DB_USERNAME_FIELD}")
 db_pass=$(${TERRAFORM} output -raw -state="${STATE_FILE_PATH}" "${TERRAFORM_DB_PASSWORD_FIELD}")
 
+# Any thoughts here?
 # There's no way to use an env variable for mysql password, so either
 # we specify on command line or we write an options file and read it 
 # with --defaults-extra-file=file_name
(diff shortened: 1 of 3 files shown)


pburkholder:
Reverted to draft while I ponder just exactly [what line 888 does](https://github.com/cloud-gov/terraform-provision/pull/1932/files#diff-1d0bbbf28914b9bc860a34ad16f16af0a6996287a6757589f82da847185a8c68R880)

ci/scripts/create-and-update-mysql.sh
```diff
@@ -0,0 +1,21 @@
...
+# Check environment variables
+db_address=$(${TERRAFORM} output -raw -state="${STATE_FILE_PATH}" "${TERRAFORM_DB_HOST_FIELD}")
+db_user=$(${TERRAFORM} output -raw -state="${STATE_FILE_PATH}" "${TERRAFORM_DB_USERNAME_FIELD}")
+db_pass=$(${TERRAFORM} output -raw -state="${STATE_FILE_PATH}" "${TERRAFORM_DB_PASSWORD_FIELD}")
+
+# Any thoughts here?
+# There's no way to use an env variable for mysql password, so either
+# we specify on command line or we write an options file and read it 
+# with --defaults-extra-file=file_name
+mysql_adm() { mysql -h "${db_address}" -u "${db_user}" -p"${db_pass}" -e "$@"; }
```
ccostino, line 16: It's not possible to use an environment variable for the password?  Isn't that what's happening here now as this is defined and configured?  Also, does there need to be a space after the password flag like the others?  I know in some cases there is no space, so thought I'd ask to double check!

```suggestion
# Any thoughts here?
# There's no way to use an env variable for mysql password, so either
# we specify on command line or we write an options file and read it 
# with --defaults-extra-file=file_name
mysql_adm() { mysql -h "${db_address}" -u "${db_user}" -p "${db_pass}" -e "$@"; }
```

ccostino (approved):
Thanks, @pburkholder!

PR #1968, opened 2025-06-04
first rotation, creating the new key

rcgottlieb:
- Rotating the cg-ecr key
-
-

## security considerations
None

### DIFF
--- a/terraform/modules/iam_user/ecr_user/outputs.tf
+++ b/terraform/modules/iam_user/ecr_user/outputs.tf
@@ -3,17 +3,17 @@ output "username" {
 }
 
 output "access_key_id_prev" {
-  value = ""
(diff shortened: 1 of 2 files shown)


jameshochadel (approved):
Thanks @rcgottlieb. I think I use this in some CSB-related pipelines, but it pulls from a shared CredHub value, so if you're planning to update it there, my use cases should be covered.

rcgottlieb:
> Thanks @rcgottlieb. I think I use this in some CSB-related pipelines, but it pulls from a shared CredHub value, so if you're planning to update it there, my use cases should be covered.

yeah part of the process, which I'm documenting for internal docs, is to update the credhub vars as well. so you should be good

PR #2158, opened 2025-07-09
Update STIG password policies for SV-235137

pburkholder:
-
- Update STIG password policies for SV-235137
-

## security considerations
Safe. Updates pw policies


ccostino (approved):
Thanks for working through and pairing on this, @pburkholder!

So far so good with these changes, they'll address some of the findings for `SV-235137`.  We verified that the parameters can be modified in the parameter groups along with their values and data type.

PR #2159, opened 2025-07-09
Add STIG option group for MARIADB_AUDIT

pburkholder:
-
-
-

## security considerations
Safe. Meets STIG baseline


ccostino (approved):
Thanks, @pburkholder!

It looks like this should do the trick, but noting here that in case it does nothing, we may have to explicitly add the option group reference to the DB resource definition itself.

PR #2160, opened 2025-07-09
Add option_group_name to database definition

pburkholder:
-
- Add Mysql (only) option group to database def
-

## security considerations
Safe - continues STIG hardening work following our standard practices.


ccostino (approved):
Thanks, @pburkholder!

PR #2179, opened 2025-07-23
This may suffice to fix the output not found error from TF

pburkholder:
-
- Downstream scripts call outputs that aren't nested deep in a module, so adding top level outputs for them to use.
-

## security considerations
Safe, fixes TF output.


cweibel (approved):
Pairing w/Peter, good to go

PR #2181, opened 2025-07-24
Just do an echo to unblock the build

pburkholder:
-
- Current run errors out with:

```
ERROR 2003 (HY000): Can't connect to MySQL server on 'development-(redacted).us-gov-west-1.rds.amazonaws.com:3306' (110)
```
- So let's not connect and just unblock the pipeline 
-

## security considerations
Safe. Only echoes static text


ccostino (approved):
Approving to unblock the pipeline!

PR #2200, opened 2025-08-12
Temporarily remove condition

jameshochadel:

I didn't write the condition correctly in 898de88c5abf64cdec5c85a964452138cba68ebe and the broker cannot currently perform any IAM operations. I will test a new condition in the console and PR that after this temporary fix.

## security considerations
Temporarily grants greater permissions to the broker.


terraform/modules/csb/iam/govcloud/csb.tf
```diff
@@ -29,11 +29,6 @@ data "aws_iam_policy_document" "brokerpak_aws_ses_govcloud" {
       "iam:TagUser"
     ]
     resources = ["*"]
-    condition {
-      test     = "StringLike"
-      variable = "aws:username"
-      values   = ["csb-aws-ses-*"]
-    }
```
markdboyd, line 36: I think this is what you want

```
    condition {
      test     = "StringLike"
      variable = "aws:username"
      values   = ["${var.stack_description}-csb"]
    }
```

soutenniza (approved):

PR #2204, opened 2025-08-14
Rotate terraform-provision stale key

rcgottlieb:
- Modified the code to add a new key
- Modified the code to use prev and curr instead of v1, v2, etc

## security considerations
None


cweibel (approved):
Pairing, good 2 go

PR #2206, opened 2025-08-21
remove action override on WAF rule

markdboyd:

- remove action override on WAF rule

## security considerations

Returning the WAF rule behavior to its default action improves security


cweibel (approved):
This is for the changed referenced in #cg-dev on 8/20/25

PR #2224, opened 2025-10-01
Metric lambda inclusion of RDS

JasonTheMain:
- add in RDS permissions in iam
- update lambda to new tag and include environment variable for account
-  Add in new metrics for RDS

## security considerations
None

### DIFF
--- a/terraform/modules/metrics_s3_ingestor/iam.tf
+++ b/terraform/modules/metrics_s3_ingestor/iam.tf
@@ -63,7 +63,7 @@ resource "aws_iam_role_policy" "lambda_tag_policy" {
         ],
         "Effect" : "Allow",
         "Resource" : [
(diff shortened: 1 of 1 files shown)


terraform/modules/metrics_s3_ingestor/iam.tf
```diff
@@ -55,6 +55,16 @@ resource "aws_iam_role_policy" "lambda_tag_policy" {
...
         ]
+      },
+      {
+        "Action" : [
+          "rds:ListTagsForResource",
+          "rds:DescribeDBInstances"
+        ],
+        "Effect" : "Allow",
+        "Resource" : [
+          "arn:${var.aws_partition}:rds:${var.aws_region}:${var.account_id}:db:*"
```
markdboyd, line 66: Could we limit this to just the prefix for brokered databases?

PR #2333, opened 2026-06-22
Update policy for external domain broker com to fix delete edge case

cweibel:
- Update policy for external domain broker com to fix delete edge case
- Part of https://github.com/cloud-gov/private/issues/2979
-

## security considerations
Tested manually in staging via AWS Console change


markdboyd (commented):
GetServerCertificate is a read-only operation, so this should be safe

soutenniza (approved):

PR #2334, opened 2026-06-23
update broker iam for RDS option group permissions

jmarcos135:
- Added the permissions for RDS option groups

## security considerations

### DIFF
--- a/terraform/modules/iam_role_policy/aws_broker/policy.tf
+++ b/terraform/modules/iam_role_policy/aws_broker/policy.tf
@@ -46,6 +46,19 @@ data "aws_iam_policy_document" "aws_broker_policy" {
 
   statement {
     actions = [
+      "rds:CreateOptionGroup",
+      "rds:DeleteOptionGroup",
+      "rds:DescribeOptionGroups",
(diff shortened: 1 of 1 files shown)


terraform/modules/iam_role_policy/aws_broker/policy.tf
```diff
@@ -44,6 +44,19 @@ data "aws_iam_policy_document" "aws_broker_policy" {
...
+  statement {
+    actions = [
+      "rds:CreateOptionGroup",
+      "rds:DeleteOptionGroup",
+      "rds:DescribeOptionGroups",
+      "rds:ModifyOptionGroup"
+    ]
+
+    resources = [
+      "arn:${var.aws_partition}:rds:${var.aws_default_region}:${var.account_id}:og:*"
```
markdboyd: Can we add a custom prefix to the option groups that we manage? And then only give permission to manage those

jmarcos135: changed to `cg-aws-broker-*`

markdboyd (approved):

PR #2336, opened 2026-06-23
update broker iam for RDS option group permissions

jmarcos135:
- Added more permissions for resource: RDS option group
-
-

## security considerations

### DIFF
--- a/terraform/modules/iam_role_policy/aws_broker/policy.tf
+++ b/terraform/modules/iam_role_policy/aws_broker/policy.tf
@@ -48,17 +48,29 @@ data "aws_iam_policy_document" "aws_broker_policy" {
     actions = [
       "rds:CreateOptionGroup",
       "rds:DeleteOptionGroup",
-      "rds:DescribeOptionGroups",
       "rds:ModifyOptionGroup",
+      "rds:ModifyDBInstance",
       "rds:AddTagsToResource",
       "rds:ListTagsForResource",
-      "rds:RemoveTagsFromResource",
+      "rds:RemoveTagsFromResource"
     ]
 
     resources = [
       "arn:${var.aws_partition}:rds:${var.aws_default_region}:${var.account_id}:og:cg-aws-broker-*"
     ]
   }
+
+  statement {
+    actions = [
+      "rds:DescribeOptionGroups"
+    ]
+
+    resources = [
+      "arn:${var.aws_partition}:rds:${var.aws_default_region}:${var.account_id}:og:*"
+    ]
+  }
+
+
 
   statement {
     actions = [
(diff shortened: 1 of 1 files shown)


terraform/modules/iam_role_policy/aws_broker/policy.tf
```diff
@@ -19,6 +19,7 @@ data "aws_iam_policy_document" "aws_broker_policy" {
     resources = [
       "arn:${var.aws_partition}:rds:${var.aws_default_region}:${var.account_id}:db:cg-aws-broker-*",
       "arn:${var.aws_partition}:rds:${var.aws_default_region}:${var.account_id}:pg:cg-aws-broker-*",
+      "arn:${var.aws_partition}:rds:${var.aws_default_region}:${var.account_id}:og:cg-aws-broker-*",
```
markdboyd: There are alot more permissions here than just for tagging. I know that some of that predates your work here, but we should at least move the `AddTagsToResource`, `ListTagsForResource`, and `RemoveTagsFromResource` permissions to a separate block that includes the options groups

markdboyd (dismissed):

terraform/modules/iam_role_policy/aws_broker/policy.tf
```diff
@@ -48,15 +48,30 @@ data "aws_iam_policy_document" "aws_broker_policy" {
...
     ]
   }
 
+  statement {
+    actions = [
+      "rds:DescribeOptionGroups"
+    ]
+
+    resources = [
+      "arn:${var.aws_partition}:rds:${var.aws_default_region}:${var.account_id}:og:*"
```
markdboyd, line 69: We should only need to describe the groups with our custom prefix, right?

jmarcos135: This needs to be * so the cleanup job can scan across all groups and check if each one is a custom broker one

terraform/modules/iam_role_policy/aws_broker/policy.tf
```diff
@@ -48,15 +48,30 @@ data "aws_iam_policy_document" "aws_broker_policy" {
     actions = [
       "rds:CreateOptionGroup",
       "rds:DeleteOptionGroup",
-      "rds:DescribeOptionGroups",
-      "rds:ModifyOptionGroup"
+      "rds:ModifyOptionGroup",
+      "rds:ModifyDBInstance",
```
markdboyd, line 52: This permission already exists above, so I don't think we need it here

jmarcos135: I added ModifyDbInstance here because it needs access to option group resources, not just db resources

markdboyd: TIL: I didn't know that `ModifyDBInstance` as a permission would apply to option groups

markdboyd (approved):

PR #2342, opened 2026-07-20
Opensearch snapshot

JasonTheMain:
- make IA so we can save 45%
-
-

## security considerations
None


terraform/modules/s3_bucket/log_encrypted_bucket/variables.tf
```diff
@@ -20,6 +20,18 @@ variable "expiration_days" {
   default = 0
 }
 
+variable "transition_storage_class" {
+  description = "Storage class to transition objects to via the lifecycle rule"
+  type        = string
+  default     = "GLACIER_IR"
```
markdboyd, line 26: Should we make the new default `STANDARD_IA` given the issues we've discovered with Glacier?

JasonTheMain: This will affect the Falco boys, not sure how long we retain it?

markdboyd: I see. I guess we can leave it as is for now then. Because if we change it then we'll have to spend money to switch from Glacier to Infrequent Access

markdboyd (approved):

PR #2344, opened 2026-07-23
s3 bucket versioning

JasonTheMain:
- allow for cleanup of non-current markers
-
-

## security considerations
None

### DIFF
--- a/terraform/modules/s3_bucket/encrypted_bucket/encrypted_bucket.tf
+++ b/terraform/modules/s3_bucket/encrypted_bucket/encrypted_bucket.tf
@@ -27,9 +27,6 @@ resource "aws_s3_bucket_acl" "encrypted_bucket_acl" {
 
 resource "aws_s3_bucket_lifecycle_configuration" "encrypted_bucket_lifecycle" {
   bucket = aws_s3_bucket.encrypted_bucket.id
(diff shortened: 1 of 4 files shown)


terraform/modules/s3_bucket/encrypted_bucket/encrypted_bucket.tf
```diff
@@ -27,22 +27,51 @@ resource "aws_s3_bucket_acl" "encrypted_bucket_acl" {
...
+      noncurrent_version_expiration {
+        noncurrent_days = 1
+      }
+
+      expiration {
+        expired_object_delete_marker = true
+      }
+
+      abort_incomplete_multipart_upload {
+        days_after_initiation = 1
```
markdboyd: So I think this might delete multipart uploads that haven't succeeded after 1 day. I'm debating if we should also wait 365 days?

markdboyd: After doing some reading on it, I think this is a good configuration. But I think we should move this `abort_incomplete_multipart_upload` block to its own rule that is not dynamic, because right now we're only applying it when bucket versioning is enabled, but deleting incomplete multipart uploads is independent of bucket versioning

markdboyd (approved):
Great work. Thanks Jason!

Issue #2348, opened 2026-08-03
Enforce Oracle TLS-only: remove plaintext 1521 ingress from rds_network SG


pburkholder:
## Context

`terraform/modules/rds_network/sg_oracle.tf` now allows ingress on **2484/tcp (TCPS)** for Oracle encryption-in-transit, enabling the broker-provisioned SSL option group's TLS listener (see aws-broker: Oracle SSL option group + TCPS binding credentials on port 2484).

To stay non-breaking, the existing **1521/tcp plaintext** ingress rule was **intentionally kept** alongside 2484. This blocked task is to remove it.

## Problem

With both 1521 and 2484 open, connections *can* use TLS but are not *forced* to. Leaving plaintext 1521 reachable is an **SC-8 (Transmission Confidentiality) finding** — a client can still connect unencrypted, and the credential travels in the clear.

## Work

- Remove the `1521` `ingress` block from `aws_security_group.rds_oracle` in `terraform/modules/rds_network/sg_oracle.tf`, leaving only the 2484/TCPS rule.
- `terraform fmt` + `terraform validate` + `plan` to confirm only the 1521 rule is destroyed.

## Blocking dependency / trigger

**Do NOT remove 1521 until all Oracle consumers connect over 2484/TLS.** Trigger for pickup: after the brokered Oracle instances and any tooling/validation paths (e.g. the cg-oracle-19c-stig-overlay `oraquery` runner using `ORAQUERY_TLS=verify-ca`) are confirmed to use 2484, and no consumer still depends on plaintext 1521.

## To Do

- [ ] Remove wordy comments left by Claude

## References
- Adds 2484 ingress: branch `feat/oracle-tls-2484-ingress` (this repo)
- Broker SSL option group / TCPS credentials: cloud-gov/aws-broker (Oracle TLS work)

> **Control Mapping:** SC-8 (Transmission Confidentiality), SC-8(1), SC-13 (Cryptographic Protection)

pburkholder:
#2351 and others resolved this.

PR #2351, opened 2026-08-05
fix(rds_network): manage Oracle SG rules as standalone aws_security_group_rule resources

wz-gsa:
> **Update (2026-08-11):** Per review discussion (Mark + Peter), this PR is now **TLS-only** —
> the plaintext **1521** ingress rule has been removed so the Oracle SG allows only verified-TLS
> **TCPS 2484**, matching what the broker implements (its binding is 2484-only; the smoke test
> opens no DB connection, so nothing consumes 1521). This closes the platform side of
> **cloud-gov/aws-broker#541** (allow 2484 / deny 1521 → only 2484) and lets the cg-stig overlay
> drop its temporary port-forwarding override.

> This PR is **separate, preventive hardening** of the AWS Oracle SG's inline-rule
> pattern (a real but latent clobber hazard; no Oracle DBs exist yet, so its blast
> radius is currently empty). The two share a lesson — hand-edited rule lists are 
> fat-finger-prone — but are different layers and different bugs.

## Summary

Refactors the Oracle RDS security group (`terraform/modules/rds_network/sg_oracle.tf`) from **inline `ingress`/`egress` blocks** to **separate `aws_security_group_rule` resources**, matching the pattern already used by `sg_postgres.tf` and `sg_mysql.tf`.

## Why

An `aws_security_group` with **inline** `ingress`/`egress` blocks makes Terraform own the security group's **entire** rule set: on every apply it reconciles the live SG to exactly the declared blocks, so any rule added out-of-band is destroyed. Oracle was the only RDS SG still using this pattern (`sg_postgres.tf`/`sg_mysql.tf` use standalone `aws_security_group_rule` resources and are unaffected).

Standalone rule resources make Terraform manage only the rules it declares, so the SG no longer clobbers rules it didn't create. This is preventive hardening: there are no Oracle RDS databases in any environment today and nothing is a member of this SG, so the change has no runtime effect now — it's done before the Oracle offering goes live so the hazard never has a chance to fire.

## Change

- `aws_security_group.rds_oracle` keeps its address, description, `vpc_id`, and tags (no SG recreate, no new SG id, no membership change).
- Three standalone rules, each `count = var.security_groups_count` sourced via `element(var.security_groups, count.index)`:
  - `oracle_ingress_tcps` — 2484/tcp (Oracle TCPS/TLS listener)
  - `oracle_ingress_plaintext` — 1521/tcp (plaintext listener)
  - `oracle_egress_default` — all-protocol egress
- Adds a native `terraform test` (`tests/oracle_sg.tftest.hcl`, mock provider) asserting the resulting rule set.

## Migration note

Inline `ingress`/`egress` sub-blocks are not independently addressable in state, so this is inherently a **destroy-inline-rules + create-standalone-rules** operation on the same SG (no lossless `moved {}` is possible). The SG resource address is unchanged, so there is no SG replacement. Reviewers should confirm the per-environment `plan` shows **only** the Oracle SG rule resources changing (no `aws_security_group.rds_oracle` replacement, no other resource affected). With no Oracle DB attached, the rule replacement is a no-op in practice.

## Verification

- `terraform fmt -check` — clean
- `terraform validate` — success
- `checkov` — 0 failures on the module (rule `description`s added, which also clears CKV_AWS_23)
- `trivy config` — 0 misconfigurations on the module
- `terraform test` (mock AWS provider) — passes (asserts 2484 + 1521 ingress, all-egress, one rule per source SG)

## Rollout

Standard per-environment promotion with plan review: dev -> staging -> production. The existing `plan-*` jobs serve as ongoing drift detection for this SG going forward.

## Related

Unblocks a clean, single-resource follow-up to enforce TLS-only by removing the plaintext 1521 rule once all Oracle consumers use 2484 (tracked in #2348).

### DIFF
--- a/terraform/modules/rds_network/sg_oracle.tf
+++ b/terraform/modules/rds_network/sg_oracle.tf
@@ -10,6 +10,15 @@ resource "aws_security_group" "rds_oracle" {
   description = "Allow access to incoming Oracle traffic"
   vpc_id      = var.vpc_id
 
+  # Declare empty inline rule sets so Terraform actively removes any rules that
+  # were previously managed inline on this SG (notably the legacy plaintext 1521
+  # ingress). Without this, migrating from inline blocks to the standalone
+  # aws_security_group_rule resources below only ADDS the new rules and leaves the
+  # old inline ones orphaned on the live SG. All actual rules are managed as the
+  # standalone resources below (TCPS 2484 + egress only).
+  ingress = []
+  egress  = []
+
   tags = {
     Name = "${var.stack_description} - Incoming Oracle Traffic"
   }
--- a/terraform/modules/rds_network/sg_oracle.tf
+++ b/terraform/modules/rds_network/sg_oracle.tf
@@ -2,34 +2,51 @@
  * Variables required:
  *   stack_description
  *   vpc_id
+ *   security_groups        (list of source SG IDs allowed to reach Oracle)
+ *   security_groups_count  (length of security_groups)
  */
 
 resource "aws_security_group" "rds_oracle" {
   description = "Allow access to incoming Oracle traffic"
   vpc_id      = var.vpc_id
 
-  ingress {
-    from_port       = 1521
-    to_port         = 1521
-    protocol        = "tcp"
-    security_groups = var.security_groups
-  }
-
-  ingress {
-    from_port       = 2484
-    to_port         = 2484
-    protocol        = "tcp"
-    security_groups = var.security_groups
-  }
-
-  egress {
-    from_port       = 0
-    to_port         = 0
-    protocol        = "-1"
-    security_groups = var.security_groups
-  }
-
   tags = {
     Name = "${var.stack_description} - Incoming Oracle Traffic"
   }
+}
+
+resource "aws_security_group_rule" "oracle_ingress_tcps" {
+  count = var.security_groups_count
+
+  description              = "Oracle TCPS/TLS listener (2484) from allowed source SGs"
+  type                     = "ingress"
+  from_port                = 2484
+  to_port                  = 2484
+  protocol                 = "tcp"
+  source_security_group_id = element(var.security_groups, count.index)
+  security_group_id        = aws_security_group.rds_oracle.id
+}
+
+resource "aws_security_group_rule" "oracle_ingress_plaintext" {
+  count = var.security_groups_count
+
+  description              = "Oracle plaintext listener (1521) from allowed source SGs"
+  type                     = "ingress"
+  from_port                = 1521
+  to_port                  = 1521
+  protocol                 = "tcp"
+  source_security_group_id = element(var.security_groups, count.index)
+  security_group_id        = aws_security_group.rds_oracle.id
+}
+
+resource "aws_security_group_rule" "oracle_egress_default" {
+  count = var.security_groups_count
+
+  description              = "Oracle egress to allowed source SGs"
+  type                     = "egress"
+  from_port                = 0
+  to_port                  = 0
+  protocol                 = "-1"
+  source_security_group_id = element(var.security_groups, count.index)
+  security_group_id        = aws_security_group.rds_oracle.id
 }
--- a/terraform/modules/rds_network/sg_oracle.tf
+++ b/terraform/modules/rds_network/sg_oracle.tf
@@ -27,18 +27,6 @@ resource "aws_security_group_rule" "oracle_ingress_tcps" {
   security_group_id        = aws_security_group.rds_oracle.id
 }
 
-resource "aws_security_group_rule" "oracle_ingress_plaintext" {
-  count = var.security_groups_count
-
-  description              = "Oracle plaintext listener (1521) from allowed source SGs"
-  type                     = "ingress"
-  from_port                = 1521
-  to_port                  = 1521
-  protocol                 = "tcp"
-  source_security_group_id = element(var.security_groups, count.index)
-  security_group_id        = aws_security_group.rds_oracle.id
-}
-
 resource "aws_security_group_rule" "oracle_egress_default" {
   count = var.security_groups_count
 
--- a/terraform/modules/rds_network/tests/oracle_sg.tftest.hcl
+++ b/terraform/modules/rds_network/tests/oracle_sg.tftest.hcl
@@ -31,6 +31,16 @@ run "oracle_sg_has_expected_rules" {
     error_message = "rds_oracle SG must exist with the expected description"
   }
 
+  # --- SG owns empty inline rule sets so legacy inline rules (e.g. 1521) are removed ---
+  assert {
+    condition     = length(aws_security_group.rds_oracle.ingress) == 0
+    error_message = "rds_oracle must declare an empty inline ingress set (rules are standalone; forces removal of legacy inline rules)"
+  }
+  assert {
+    condition     = length(aws_security_group.rds_oracle.egress) == 0
+    error_message = "rds_oracle must declare an empty inline egress set (rules are standalone)"
+  }
+
   # --- 2484 TCPS ingress rule (the only ingress — TLS-only) ---
   assert {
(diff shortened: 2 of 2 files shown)


terraform/modules/rds_network/sg_oracle.tf
```diff
@@ -2,34 +2,51 @@
...
+  to_port                  = 2484
+  protocol                 = "tcp"
+  source_security_group_id = element(var.security_groups, count.index)
+  security_group_id        = aws_security_group.rds_oracle.id
+}
 
-  tags = {
-    Name = "${var.stack_description} - Incoming Oracle Traffic"
-  }
+resource "aws_security_group_rule" "oracle_ingress_plaintext" {
```
markdboyd, line 30: Why is this named `oracle_ingress_plaintext`? Would `oracle_ingress_default` be a better name since 1521 is the default port for Oracle?

wz-gsa: I wanted it to be obvious that it was plaintext so we don't enable transport over it. oracle_ingress_default is probably fine but I wanted to err on the side of making non TLS traffic obvious. 

@markdboyd what do you think?

markdboyd: @wz-gsa Why can't we or don't we enable TLS trafic over port 1521? Isn't that the default for Oracle? Is it recommended not to use 1521?

wz-gsa: Good question — it gets at the exact reason the naming matters.

**1521 and 2484 aren't one port with a TLS toggle; they're separate listener endpoints with different protocols.** By Oracle convention 1521 is `PROTOCOL=TCP` (plaintext) and 2484 is `PROTOCOL=TCPS` (TLS). I confirmed against the merged broker code (`aws-broker` `services/rds/oracle_tls.go`): the Oracle SSL option group opens TLS specifically on **2484** (`oracleSSLPort = 2484`, `SQLNET.SSL_VERSION=1.2`, `FIPS.SSLFIPS_140=TRUE`). So on our RDS instances, a connection over 1521 is unencrypted *by definition* — it isn't that we've disabled TLS on 1521, it's that 1521 is the non-TLS endpoint and 2484 is the TLS one.

**On "why can't we just turn off 1521":** we can't do it from the broker, and RDS doesn't expose a switch to close the base TCP listener — it's always present. The broker code says this explicitly: *"the broker attaches the SSL option and opens 2484, but cannot deny plaintext 1521 — that security-group rule is owned by the platform (cg-provision/Terraform), outside the broker's IAM."* In other words, **the only lever to block plaintext 1521 is a security-group deny — i.e. this repo.** Removing the 1521 ingress rule here is that lever, tracked as the TLS-only follow-up (#2348 / aws-broker#541).

**Correction on why 1521 is retained right now:** to be accurate — this is a **new, not-yet-launched offering, so there are no customer consumers on 1521.** The only things that could use 1521 today are our own dev/CI validation (the runner harness + broker smoke tests), and even those default to verified TLS on 2484. So keeping 1521 is just sequencing hygiene while the broker binding/smoke path is confirmed fully 2484-only — not backward-compat for an installed base. Which means the TLS-only cutover (dropping 1521) is low-risk and can happen before GA; there's no customer migration gating it.

That's why I named it `oracle_ingress_plaintext` rather than `oracle_ingress_default`: the SC-8-relevant fact about that rule is that it carries *cleartext* traffic and is the one we intend to drop — "default" would hide exactly that. If you'd prefer consistency with `oracle_ingress_tcps`, I'm happy to rename it to **`oracle_ingress_tcp`** (TCP vs TCPS reads cleanly as the plaintext/TLS pair). Your call.

markdboyd: @wz-gsa If we are only ever going to allow customers to access Oracle DBs over port 2484, and that's what the broker code implements, then I think we should remove the ingress rule for port 1521

pburkholder: That would also close https://github.com/cloud-gov/aws-broker/issues/541

wz-gsa: Agreed — will do. Making #2351 **TLS-only**: removing the plaintext 1521 ingress rule so the Oracle SG allows only TCPS 2484, matching what the broker actually implements (verified: `credentials.go` emits a TCPS-only binding on 2484; the smoke test doesn't open a DB connection; no consumer of the 1521 rule). I'll update the `terraform test` + PR description and sync the branch with `main`.

This closes **aws-broker#541** (allow 2484 / deny 1521 is now just "only 2484") and lets the cg-stig overlay drop its temporary port-forwarding override. Pushing shortly.

markdboyd (dismissed):

wz-gsa:
Not in dev yet — to be precise about what's been validated:

**Done:** static checks pass — Concourse `format-terraform` + `validate-terraform`, CodeQL/Analyze, and a local `terraform test` (mock provider) asserting the resulting rule set (TCPS 2484 ingress + egress, no 1521). Commits are now GPG-signed with the Concourse-trusted key.

**Not done:** this hasn't been through `plan-development` / `apply-development` — those run on `main` post-merge, so the PR branch hasn't been planned against dev.

**Why it should be low-risk regardless** (from inspecting current state, read-only): the Oracle SG's only ingress source is the bosh SG, and there are **no Oracle RDS instances in any environment**, so nothing is a member of this SG. The expected dev plan is therefore just *"destroy the two `oracle_ingress_plaintext` (1521) rule resources, no other resource affected,"* with zero runtime impact since no DB is attached. (The broker binding is 2484-only and the smoke test opens no DB connection, so nothing consumes 1521.)

If you'd like an actual dev plan before merge, happy to coordinate running `plan-development` against this branch (or hold while someone triggers it). Otherwise it'll show up in the dev plan on merge. Your call on which you'd prefer.

wz-gsa:
Confirmed your suspicion, Mark — and fixed it. Read the build 6817 plan: `2 to add, 1 to change, 0 to destroy`; the 2 adds were the new standalone `oracle_ingress_tcps` (2484) + egress **on the existing SG `sg-0db2c5bad147b9aec`** (so it wasn't a *new* SG — the `rds_oracle` SG itself wasn't in the plan; the "1 to change" is the unrelated aws-broker IAM policy). The problem was exactly what you flagged: **nothing removed the legacy inline 1521 rule.**

Root cause: the refactor moved rules from inline `ingress{}` blocks to standalone `aws_security_group_rule` resources. The AWS provider does not reconcile inline→standalone automatically — it just adds the new rules and leaves the old inline ones orphaned. A live describe confirmed the SG currently has **both** inline rules (1521 **and** 2484 from `sg-2b4ddc4e`), so 1521 would have survived — TLS-only not actually achieved.

Fix (pushed, `175212aa`): declare empty inline sets (`ingress = []` / `egress = []`) on `aws_security_group.rds_oracle` so Terraform actively **removes** the previously inline-managed rules; the standalone resources then add back only 2484 + egress. The next `plan-development` should now show the SG **updated in-place to strip the inline 1521/2484 rules** (net live result: 2484 + egress only), rather than 0-to-destroy.

Could you (or I, if you point me at a dev role I can assume — my `gov-dev-*` profiles hit AssumeRole AccessDenied) re-run `plan-development` to confirm it now shows 1521 being removed? One caveat: with `ingress=[]` + standalone rules, some AWS provider versions can show a perpetual diff; if the re-plan shows that, the fallback is a one-time `terraform state` removal of the inline rules instead — but the standard expectation is a clean one-time in-place strip.

pburkholder (approved):
The plan looked good so I ran the apply too. The Cloud.gov app can still connect over port 2484.

pburkholder:
@wz-gsa yeah, my comment at https://github.com/cloud-gov/terraform-provision/pull/2351#pullrequestreview-4917325118 was probably insufficient. The CI run at https://ci.fr.cloud.gov/teams/main/pipelines/terraform-provision/jobs/apply-development/builds/1237 applied exactly what we expected, so you can go ahead merge.

PR #2355, opened 2026-08-13
Logsearch retirement

JasonTheMain:
- This is for the retirement of logsearch
- Retirement of resources associated with logsearch
-

## security considerations
Less buckets and resources to manage


terraform/modules/bosh_vpc/sg_bosh.tf
```diff
@@ -204,7 +195,7 @@ resource "aws_security_group_rule" "monitoring_elasticsearch_exporter" {
   security_group_id = aws_security_group.bosh.id
 }
 
-resource "aws_security_group_rule" "concourse_logsearch" {
+resource "aws_security_group_rule" "concourse_opensearch" {
```
markdboyd: Do we need this rule? Do we allow concourse to talk to OpenSearch on port 9200?

JasonTheMain: That is the opensearch port

markdboyd (approved):

PR #2358, opened 2026-08-13
add scanning users for databases

dandersonsw:
- Creates database users for running our scans

## security considerations
Having separate users keeps the blast radius down if there is ever any compromise of accounts

### DIFF
--- a/ci/scripts/create-and-update-db.sh
+++ b/ci/scripts/create-and-update-db.sh
@@ -59,9 +59,9 @@ for db in ${DATABASES}; do
 			        ALTER ROLE ${db_scanner_user} WITH PASSWORD '${db_scanner_pass}';
 			      END IF;
 
-			      GRANT USAGE ON SCHEMA public TO ${db_scanner_user}";
-			      GRANT SELECT ON ALL TABLE IN SCHEMA public TO ${db_scanner_user}";
-			      GRANT pg_read_all_settings TO ${db_scanner_user}";
+			      GRANT USAGE ON SCHEMA public TO ${db_scanner_user};
+			      GRANT SELECT ON ALL TABLE IN SCHEMA public TO ${db_scanner_user};
+			      GRANT pg_read_all_settings TO ${db_scanner_user};
 			    COMMIT;
 		EOT
 	fi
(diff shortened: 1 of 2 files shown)


ci/scripts/create-and-update-db.sh
```diff
@@ -19,98 +21,112 @@ psql_adm() { psql -h "${db_address}" -U "${db_user}" "$@"; }
...
+			      IF NOT EXISTS (SELECT 1 FROM pg_roles WHERE rolname = '${db_scanner_user}') THEN
+			        CREATE ROLE ${db_scanner_user} WITH LOGIN PASSWORD '${db_scanner_pass}';
+			      ELSE
+			        ALTER ROLE ${db_scanner_user} WITH PASSWORD '${db_scanner_pass}';
+			      END IF;
 
-  # Remove default privileges
-  psql_adm -d "${db}" -c "REVOKE ALL ON SCHEMA public FROM PUBLIC"
+			      GRANT USAGE ON SCHEMA public TO ${db_scanner_user};
+			      GRANT SELECT ON ALL TABLE IN SCHEMA public TO ${db_scanner_user};
```
cweibel: GRANT SELECT ON ALL TABLES IN SCHEMA schema_name TO username;

think TABLES should be plural

cweibel: Also,  GRANT SELECT ON ALL TABLES ... only affects tables that exist at run time. Tables created later won't be readable by the scanner unless you also add ALTER DEFAULT PRIVILEGES ... GRANT SELECT ... TO scanner.

ci/scripts/create-and-update-db.sh
```diff
@@ -19,98 +21,112 @@ psql_adm() { psql -h "${db_address}" -U "${db_user}" "$@"; }
...
-  psql_adm -d postgres -l | awk '{print $1}' | grep -q "${db}" || \
-    psql_adm -d postgres -c "CREATE DATABASE ${db} OWNER ${db_user}"
-  # Enable extensions
-  for ext in ${EXTENSIONS}; do
-    psql_adm -d "${db}" -c "CREATE EXTENSION IF NOT EXISTS \"${ext}\""
-  done
+	# Create scan user if it does not exist and grant permissions needed for scans
+	if [ -n "${db_scanner_user}" ] && [ -n "${db_scanner_pass}" ]; then
+		psql_adm -d "${db}" <<-EOT
+			    BEGIN;
```
cweibel, line 55: IF ... THEN ... END IF is PL/pgSQL, not plain SQL. You cannot use IF NOT EXISTS (...) THEN at the top level of a psql script. That syntax is only valid inside a DO $$ ... $$ block or a function body. As written, psql will throw a syntax error at IF. (Contrast with the existing f_enforceCloudGovOrigin code, which correctly wraps procedural logic in $$ ... $$ LANGUAGE plpgsql.)

cweibel (changes requested):

PR #2361, opened 2026-08-26
fix(rds_network): exclude tooling from Oracle SG rules

pburkholder:
- Add an `oracle_rules_enabled` bool (default `false`) to `modules/rds_network`
- Ensure it's enabled in main (dev/staging/prod)
- Ensure it's false in tooling
- Add a couple of tests.

## Verification

| `terraform test` — `modules/rds_network` (mock provider) | **3 passed, 0 failed** |

## Security considerations

Improves default assumptions, requires explicitly setting Oracle ingress/egress.
---
🤖 This PR was prepared with AI assistance (OpenCode). All changes require human review before merge; the plan output above has not been validated against live AWS state.

### DIFF
--- a/terraform/modules/rds_network/sg_oracle.tf
+++ b/terraform/modules/rds_network/sg_oracle.tf
@@ -4,7 +4,14 @@
  *   vpc_id
  *   security_groups        (list of source SG IDs allowed to reach Oracle)
  *   security_groups_count  (length of security_groups)
+ *   oracle_rules_enabled   (false in stacks with no Oracle RDS, e.g. tooling)
  */
+
+locals {
+  # Rule count per source SG, gated by oracle_rules_enabled so stacks without an
+  # Oracle RDS (tooling) create no Oracle rules at all.
+  oracle_rules_count = var.oracle_rules_enabled ? tonumber(var.security_groups_count) : 0
+}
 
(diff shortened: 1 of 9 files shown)


terraform/modules/rds_network/sg_oracle.tf
```diff
@@ -2,10 +2,16 @@
...
- *   security_groups        (list of source SG IDs allowed to reach Oracle)
- *   security_groups_count  (length of security_groups)
+ *   security_groups           (list of source SG IDs allowed to reach Oracle)
+ *   security_groups_count     (length of security_groups)
+ *   rds_oracle_rules_enabled  (false in stacks with no Oracle RDS, e.g. tooling)
  */
 
+locals {
+  # Zero rules unless var.rds_oracle_rules_enabled is True (e.g. false in Tooling)
+  oracle_rules_count = var.rds_oracle_rules_enabled ? var.security_groups_count : 0
```
markdboyd, line 12: There is nothing wrong with this approach, but we could simplify by just having a variable named `oracle_security_groups_count` and then make the statements on resources below be

```
count = var.oracle_security_groups_count
```

Then you wouldn't need this extra `rds_oracle_rules_enabled` boolean variable that you have to pass down from other stacks

markdboyd (approved):

PR #2363, opened 2026-08-31
rotating out old key

rcgottlieb:
- rotating the IAM keys step 2: remove old key

## security considerations
None


cweibel (approved):
This is fine to run as long as any credhub or similar references to the old key have been removed downstream

PR #2364, opened 2026-09-01
fix tagging permissions for broker 

markdboyd:

- fix tagging permissions for broker to allow tagging on parameter groups

## security considerations

These permissions are necessary for the broker's normal operation when creating and deleting parameter groups. The permissions are scoped to only the resources that the broker is managing


JasonTheMain (approved):
LGTM

PR #2367, opened 2026-09-02
Downgrade RDS DB engine version for STIG

pburkholder:
Updated RDS DB engine version to comply with cybersecurity standards and added description for the variable.

-
-
-

## security considerations
Complies.

### DIFF
--- a/terraform/modules/csb/broker/variables.tf
+++ b/terraform/modules/csb/broker/variables.tf
@@ -26,13 +26,19 @@ variable "rds_db_engine" {
 }
 
 variable "rds_db_engine_version" {
+  description = <<-EOT
+    Per Cloud.gov cybersecurity, all MySQL DBs within our boundary must
+    adhere to STIG standards. Per stighub.com and stigviewer.com, only MySQL 8.0
+    has published standars as of 2026-09-02
+  EOT
   type    = string
-  default = "8.4"
+  default = "8.0"
 }
 
 variable "rds_parameter_group_family" {
+  desc
   type    = string
-  default = "mysql8.4"
+  default = "mysql8.0"
 }
 
 variable "rds_username" {
--- a/terraform/modules/csb/broker/variables.tf
+++ b/terraform/modules/csb/broker/variables.tf
@@ -29,14 +29,13 @@ variable "rds_db_engine_version" {
   description = <<-EOT
     Per Cloud.gov cybersecurity, all MySQL DBs within our boundary must
     adhere to STIG standards. Per stighub.com and stigviewer.com, only MySQL 8.0
-    has published standars as of 2026-09-02
+    has published standards as of 2026-09-02
   EOT
-  type    = string
-  default = "8.0"
+  type        = string
+  default     = "8.0"
 }
 
 variable "rds_parameter_group_family" {
-  desc
   type    = string
   default = "mysql8.0"
 }


markdboyd (dismissed):

markdboyd (approved):

wz-gsa (approved):
**APPROVE.** The change is correct and the direction is right. Two things worth recording, one of which explains why #2366 probably needed reverting at all.

### Verified
- The defaults are **live**, not dead config: `stacks/main/stack.tf:522` is the only `csb_broker` invocation and it passes neither `rds_db_engine_version` nor `rds_parameter_group_family`, so both defaults flow through `csb/broker/main.tf` → `modules/rds_stig`. No tfvars or `TF_VAR_` override exists in `ci/pipeline.yml` (only `csb_rds_password`).
- **Internally consistent** — no `8.4`/`mysql8.4` left anywhere in the repo. The other `rds_db_engine_version` defaults (12.17, 15.7) are postgres modules and are untouched.
- `major_engine_version = var.rds_db_engine_version` in `rds_stig/option_group.tf` also becomes `8.0`, so engine / parameter-group family / option-group major version now agree. Under #2366 they were consistent at 8.4 too, so this isn't a fix of a mismatch — just noting the three move together.
- The STIG rationale holds up against the repo: `rds_stig/parameter_group.tf` already implements MySQL STIG parameters and cites a V-ID (`# SV-235137`), which is from the **MySQL 8.0** benchmark. So the module's existing content corroborates the 8.0 justification independently of the description's citation.

### FIX — the description cites third-party mirrors as the authority
```
Per stighub.com and stigviewer.com, only MySQL 8.0
has published standards as of 2026-09-02
```
These are mirrors. For a version pin whose stated justification is compliance, the citation should be the DISA source (`public.cyber.mil/stigs/downloads` — "Oracle MySQL 8.0 Server STIG"), with the mirrors as convenience links if wanted. An assessor tracing "why is this pinned to 8.0" should land on the authoritative catalogue. Also worth adding the benchmark title/version rather than a bare date, since "as of 2026-09-02" goes stale silently.

### FYI, no action — why #2366 likely failed rather than merely being undesirable
`rds_stig/database.tf:39` sets `allow_major_version_upgrade = var.rds_allow_major_version_upgrade`, and `stacks/main/variables.tf:147` defaults it to **`false`**. MySQL 8.0 → 8.4 is a **major** version change. If the CSB instance was already deployed at 8.0 (the default before #2366), an apply of #2366 would have been rejected by RDS rather than upgrading — so this revert plausibly unblocks a broken apply, not just a policy preference.

Two consequences worth knowing:
1. **If no apply ran between #2366 merging (17:59Z) and this PR, #2367 is a no-op against real infrastructure** — the instance is still on 8.0 and the plan is clean. If an apply *did* run and somehow succeeded, this becomes a **downgrade**, which RDS does not support in place; `engine_version` is not in the module's `ignore_changes` (only `identifier` is), and `aws_db_parameter_group.family` is ForceNew, so the parameter group and option group would be replaced. I can't see Concourse from here — worth confirming no `apply-*` job ran in that window before merging.
2. Whenever MySQL 8.4 (or 9.x) does get a published STIG, moving to it will need `rds_allow_major_version_upgrade = true` for that apply. Worth a note in the description so the next person doesn't repeat #2366.

### FYI — PR body
`## Changes proposed in this pull request:` is three bare dashes, and `## security considerations` is "Complies." For a change whose entire rationale *is* the security consideration, one line naming the STIG and the reverted PR would carry the reasoning into the git record. Recent PRs (#2363-#2365) do fill these in.

Nothing blocking.

PR #2368, opened 2026-09-04
Bump to MySQL to 8.4

pburkholder:
-
- bump to 8.4
- creates new option groups and parameter groups before deleting the existings
- TF's attempt to delete the old option group on the first apply will fail, since that option group is still associated with the automatic snapshots
   - After the upgrade, we have to manually delete the snapshots, then re-apply

## security considerations
Making sure we don't break STIG validation....

### DIFF
--- a/terraform/modules/rds_stig/option_group.tf
+++ b/terraform/modules/rds_stig/option_group.tf
@@ -10,12 +10,22 @@ resource "aws_db_option_group" "option_group_mysql" {
     option_name = "MARIADB_AUDIT_PLUGIN"
   }
 
-  # DB snapshots record the option group they were taken with, so automated
-  # backups pin the superseded group until they age out of
-  # backup_retention_period (35 days, see database.tf). Leave it in place;
-  # deleting it fails with InvalidOptionGroupStateFault even once the instance
-  # has moved to the new group. The orphan is free but is not auto-cleaned.
-  skip_destroy = true
+  # Regarding major version updates:
+  # DB snapshots record the option group they were taken with, so
+  # backups reference the old option group until they age out of
+  # backup_retention_period (35 days, per database.tf). Attempting to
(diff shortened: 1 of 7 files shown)


terraform/modules/rds_stig/option_group.tf
```diff
@@ -13,4 +9,28 @@ resource "aws_db_option_group" "option_group_mysql" {
...
+  # backups reference the old option group until they age out of
+  # backup_retention_period (35 days, per database.tf). Attempting to
+  # delete the option group before then
+  # fails with InvalidOptionGroupStateFault
+  # ==========
+  # HOWEVER, we do not use `skip_destroy = true`
+  # because we don't want the option group to remain around forever.
+  # Instead, we
+  # - Do the major version update with TF (it will fail to delete the old option group)
+  # - Connect to the CSB to run
```
markdboyd, line 24: This module could be used for other databases than the CSB. Also, is a local database backup necessary if we do a manual snapshot?

pburkholder: Yes, this is going to be too much work to safely maintain if we ever having to upgrade more than one STIG'd DB. 

We could just keep the option_group around forever by having skip_destroy. It doesn't cost us anything 

We could do something with time stamps, but it seems hokey

```
skip_destroy = (now() > 2026-10-31)
```

Or we could count the number of DB snapshots that still reference that option group, and pass in a variable into the pipeline?

As for the local DB dump, I was doing a belt-and-suspenders thing based off a suggestion from Weibel. 

What are your thoughts?

markdboyd: Let's just change the text here to "Connect to the database locally and create a backup" just to make it non-CSB specific, then I think it's fine

markdboyd (approved):

